PR #268, opened 2015-12-09
Changed case on GoCD to preferred format

kmugrage:


mahendrakariya:
Thanks a lot for this PR!

PR #270, opened 2015-12-10
A more meaningful message when an env directory isn't found

haroon-sheikh:


kashishm:
Thank you for your contribution @haroon-sheikh, we appreciate it!

PR #286, opened 2016-01-14
Tidy up install script

jean:
The only functional change is to set `PATH` and `GAUGE_ROOT` independently.


mahendrakariya:
Thanks a lot @jean!!

PR #288, opened 2016-01-17
add a usage message

jean:


mahendrakariya:
Thanks a lot @jean! Much appreciated! :)

jean:
Welcome :-)  I didn't do anything about OS X

mahendrakariya:
That's fine since `install.sh` is not used on OS X. It's only for Linux distros.

Issue #175, opened 2015-07-09
Multiline step arguments


tlempart:
Support multiline step arguments similar to https://cucumber.io/docs/reference#doc-strings. It would be great to allow using different formatting options, e.g.

```
* step 1
'''gauge-parameter <- backtick
multiline step argument as code block
'''
* step 2
```

or:

```
* step 1
multiline step argument as paragraph
* step 2
```

I am not sure about the second example, because sometimes there is a need to document a step in specification, but in that case you can simply ignore such parameter in step implementation.

uuunk:
+1

FaustXVI:
+1. I'll try to make a PR for it.

mahendrakariya:
That'd be really awesome @FaustXVI !!

I'd also suggest you to discuss the changes that you are planning to make, before you actually start implementing it. Because the changes for this feature may not be just limited to Gauge core. It may impact other plugins as well.

FaustXVI:
Looking at the code, I think I'll just change the parser to accept multi-line strings. So :
```
* a step with "a parameter"
```
and
```
* a step with
"""
a parameter
"""
```
would be equivalent.

What do you think ?
What about the triple quotes ?

sriv:
Update, with support for #265 coming up, I believe something like this should be possible:

```
* The word "gauge
in multiple lines" 
has "9" vowels.
```

Will be the same as :

```
* The word "gauge in multiple lines" has "9" vowels.
```

Feedback welcome!

haroon-sheikh:
@FaustXVI @sriv Do we have any updates on this? As part of this we should look at using code blocks to allow passing of `json/xml` strings in the parameters.

E.g.
```json
{"hello":"world"}
```

sriv:
@haroon-sheikh - #265 brings support for mulitline step text (although it is toggled off, by default)

Multiline step arguments needs some work. #1240 also talks about code parameters which I believe is similar to what you seek. Perhaps you could add your thoughts there?

zabil:
To make this usable within markdown syntax it's best to use code blocks as @haroon-sheikh mentions.

For example (provided by @ajoecker)

````
* When sending 
```
{
  human(id: "1000") {
    name
    height
  }
}
```
*  Then the name is "Batman"
````

This means multiline parameters can only be used at the the end of a step in a separate line.
This will also partially meet the requirements of https://github.com/getgauge/gauge/issues/1240

zabil:
[...]

Issue #290, opened 2016-02-01
Export specs to HTML


mahendrakariya:
Users should be able to export the specs to HTML.
For more details, refer https://groups.google.com/forum/#!topic/getgauge/eDb-GAVXYjg

bensullivan:
It would be good if this feature allowed specs to be searched, toggling of test execution time and result (so that you can just focus on the behaviour if you want to) and the grouping of specs, similar to Topics in relishapp.

Many thanks

sriv:
[Tech Notes]

Options evaluated:

- Golang markdown libraries such as [blackfriday](https://github.com/russross/blackfriday) may not suit this requirement, since blackfriday omits tags it does not understand, as a result, comments and `tags` are left out. One option is to fork and maintain a version that is more liberal and suit's gauge's requirement.
Pros: can have a native binary making deployment/setup simple.
Cons: Loss of relevant comments/tags

- look at [jekyll](http://jekyllrb.com), and have a script/plugin that will generate a site based on the theme selected. Jekyll structure is different from gauge's convention, and requires YAML Front Matter for meta information (although optional).
Pros: Very popular, plenty of themes available.
Cons: Requires conversion from gauge to jekyll structure. Ruby needs to be installed on users machine.

- Use [mkdocs](http://www.mkdocs.org/)
Pros: Simplest to use to generate documentation. `readthedocs` theme is beautiful.
Cons: Requires conversion from `spec`, `cpt` to `md` file extension (mkdocs works only with known markdown extensions). Python and pip needs to be installed on users machine to install mkdocs.

Of these, I prefer using mkdocs.

kashishm:
We have released a [plugin](https://github.com/getgauge/M2H/releases/tag/v0.0.1) which provides a basic `Export to HTML` functionality with filtering by tag expression. Refer [readme](https://github.com/getgauge/M2H/blob/master/README.md) for documentation. Please log an issue [here](https://github.com/getgauge/M2H/issues) if you have any feedback.

Issue #265, opened 2015-12-02
Support newline in step text


falkoschumann:
When I insert a newline into long step text, Gauge use only the first line as step definition and ignore the second line.

Example:

```
* Step with a long
  line.
```

Gauge use only "Step with a long" as step definition and ignore "line."

I use Gauge with the Java plugin.

sriv:
This is by design, Gauge's grammar expects a step text to begin with an asterix (*) and end with a Newline. 

Would having a single line text, that can be word-wrapped in your editor be an option worth considering?

If Gauge has to support this, One way would be to introduce a way to escape newline in Gauge. Something like this:

```
* Step with a very very long line \
  that flows into two lines
```

Thoughts?

falkoschumann:
Soft wrap does not intend the wrapped line. It's only a small optical lack for printing as plain text for customer review. Specifications are written in Markdown so I expected the newline behavior of Markdown:

>To make lists look nice, you can wrap items with hanging indents [1]

When I write Markdown, escape the newline feels not good for me. Export or print the existing HTML preview of specification in IntelliJ would be a alternative to using plain text for communicating with customers. For the moment soft wrap is O.k.

[1]: https://daringfireball.net/projects/markdown/syntax#list

Thunderforge:
To clarifying what @falkoschumann is saying, the issue is that Gauge handles line breaks differently [than standard Markdown does](http://daringfireball.net/projects/markdown/syntax#p). In standard Markdown, a single break return after a list item is rendered as follows:

**Markdown:**

    * Asterisk
    No asterisk
    * Asterisk

**GitHub Render:**

* Asterisk
No asterisk
* Asterisk

**Gauge HTML Preview Render:**

*This renders it as though it had an intentional `<br/>` in it.*

![image](https://cloud.githubusercontent.com/assets/6200170/11544610/6e6b5d04-9908-11e5-9016-bd3df1689db0.png)

This is not how Gauge behaves. Instead it treats `No asterisk` as a comment distinct from the line above it.

Moreover, Gauge's HTML Preview tab (in IDEA) renders it as though it had an intentional `<br/>` in it:

This is the issue: Gauge doesn't reflect standard Markdown:

-------

As for a solution, I don't like the idea of adding a `\` to the end of the line, as that just makes Gauge deviate further from standard Markdown. An ideal solution would be to just require comments to be outside of a list. Not only is this proper Markdown, but it already renders properly in Gauge HTML Preview:

**Markdown:**

[...]

Issue #172, opened 2015-07-07
Alternative way to quote parameters


tlempart:
Especially in case of _JSON_ values there is a need to use a large number of double-quotes `"`, e.g.: `{ "parameter1": "value1", "parameter2": "value2" }. If I want to use such *JSON* as parameter, than I have to escape every one`"` character. It would very useful to have possibility to use parameter quotes other than double-quotes, e.g. single quotes.

deepthigc:
One way to do this would be to put this JSON segment in a file and use the (file parameter)[http://getgauge.io/documentation/user/current/specifications/parameters.html] that Gauge supports.

Changing the parameters to support single quotes is one way to achieve this, but I feel having JSON embedded within a step takes away the readability aspect that markdown tries to bring in. 

Thoughts?

hakito:
It would be great if a string quoting like in ruby would be possible.

For example:

```
* I am a step that requires json %#{"key": "value"}# data
* I am a step that requires an array %w[with 3 values] as data
```

tk8817:
I feel like this would be easily solved through the use of regex's similar to cucumber. It would allow for the capturing of data in an un-opinionated fashion.

zabil:
Introducing JSON, string syntax, regexes affects the readability of specification. 
Gauge recommends using special parameters or custom parameters https://github.com/getgauge/gauge/issues/7

PR #329, opened 2016-04-16
[WIP] Add multiline parameter capacity

FaustXVI:
First try to have multiline paramter (see #175).
**This is defintly not the final version** but I'd like to have your feedback.

TODOs 
- [ ] Quotes in mulitline should not have to be escaped
- [ ] For now, quotes needs to be on the first line. Starting it on a new line would be great
- [ ] I used two quotes instead of triple. I do think triple are nicer but it was simpler to do with double.
- [ ] tests in more a real case
- [ ] clean up some code :D

### DIFF
--- a/parser/specparser.go
+++ b/parser/specparser.go
@@ -94,6 +94,11 @@ func (parser *SpecParser) GenerateTokens(specText string) ([]*Token, *ParseError
 			newToken.Kind = gauge.ScenarioKind
 			parser.tokens = append(parser.tokens[:len(parser.tokens)-1])
 		} else if parser.isStep(trimmedLine) {
+			if parser.isMultiline(trimmedLine) {
+				for nextLine, hasNextLine := parser.nextLine(); hasNextLine && !parser.isMultiline(strings.TrimSpace(nextLine)); nextLine, hasNextLine = parser.nextLine() {
+					trimmedLine += "\n" + strings.TrimSpace(nextLine)
+				}
+			}
 			newToken = &Token{Kind: gauge.StepKind, LineNo: parser.lineNo, LineText: strings.TrimSpace(trimmedLine[1:]), Value: strings.TrimSpace(trimmedLine[1:])}
 		} else if found, startIndex := parser.checkTag(trimmedLine); found {
 			newToken = &Token{Kind: gauge.TagKind, LineNo: parser.lineNo, LineText: line, Value: strings.TrimSpace(trimmedLine[startIndex:])}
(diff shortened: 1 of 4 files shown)


sriv:
@FaustXVI - Thanks for getting this in motion. Quick question - I notice that the syntax for multiline arguments require two quotes? i.e. something like this:

```
* this is a step with ""multiline
  arguments""
```

Instead, it would be good if we can not overload parameter definition and use single double quotes, something like:

```
* this is a step with "multiline
  arguments"
```

Note that we might have to honour the indentations. There is a separate issue to support multiline step texts (#265). From [Markdown Specs](http://spec.commonmark.org/0.25/#list-items):

> If a sequence of lines Ls constitutes a list item according to rule #1, #2, or #3, then the result of indenting each line of Ls by 1-3 spaces (the same for each line) also constitutes a list item with the same contents and attributes. If a line is empty, then it need not be indented.

Example: http://spec.commonmark.org/0.25/#example-245

PS - I see that you had also [raised this](https://github.com/getgauge/gauge/issues/175#issuecomment-205003941) in #175. Apologies for not responding earlier.

sriv:
Some thoughts on your TODOs:

> Quotes in mulitline should not have to be escaped

Actually, quotes on regular static parameters are supposed to be escaped, it would be good to keep this consistent on multiline parameters as well.

Note that there is another open issue (#172) that seeks to allow alternative quotes support. That would allow users to not have to escape quotes.

> For now, quotes needs to be on the first line. Starting it on a new line would be great

Actually, I like it being continuous. When we implement #265, parameter starting on new line could be just a special case of multiline step text. What do you say?

FaustXVI:
@sriv thanks for the feedback.
Actually, I think that fixing #265 will give us multiline arguments for free and so, indeed, a single " would be nicer. I think I'll switch my focus to #265.
I do agree with #172 too since it totally solve the use case I have in mind. I'll add a comment to it.
Maybe the #175 could be marked as a kind of duplicate of #265 and #172 ?

sriv:
@FaustXVI  - cool! Have you made any progress on #265 ? We were looking at #95 and it seems to be related as well.

If you haven't made any significant changes, I would like to address #95, since there are other dependant issues on #95 (ex. #290 ).

FaustXVI:
@sriv I haven't done any significant work. Please go ahead with #95. I'll start to work on #265 in two weeks after my holydays.

mahendrakariya:
@FaustXVI The changes in this PR has diverged a lot from the master branch. I am closing it for now.

flyerfish:
has gauge multiline parameter capacity? how to write multiline parameter of step? thanks.

apoorvam:
@flyerfish Currently, Gauge does not have multi-line step text/parameter support. [Here](https://github.com/getgauge/gauge/issues/175) is an open issue for multi-line argument support. You can add your comments/use case in this issue.

Gauge allows only table parameter to be present in line after step text. Example [here](https://github.com/getgauge/gauge-example-java/blob/ed0e38e0f9a22f28239da69e618a37af22204ab8/specs/products.spec).

Issue #418, opened 2016-05-27
"gauge --failed" runs all specs from "specs/" subdirectory when `FailedItems` is empty in .gauge/failed.json


kaustavdm:
### Expected behavior

`gauge --failed` should attempt to run only failed scenarios from previous run. If `FailedItems` is empty in `.gauge/failed.json`, it should not run scenarios or specifications.

### Actual behavior

`gauge --failed` currently runs all the specs in the `specs/` subdirectory in the project root even when `FailedItems` is empty. That `specs/` directory seems to be hard-coded because `gauge --failed` does not accept paths to specs. So, you get the same behaviour on running `gauge --failed <path-to-other-specs-dir>`.

### Steps to reproduce
1. Run `gauge --failed` on a test suite with no previous failures.

### Gauge version

Run gauge -v on your system and paste the results here.

```
Gauge version: 0.4.1.nightly-2016-05-27

Plugins
-------
html-report (2.1.1)
java (0.4.1)
js (1.2.0)
ruby (0.2.0)
xml-report (0.1.2)
```

sguptatw:
gauge --failed . from a subfolder is giving the following error.
```
Failed to read last run information. Reason: File .gauge/failures.json doesn't exist.
2016/06/07 13:38:21 Failed to read last run information. Reason: File .gauge/failures.json doesn't exist.
```

mahendrakariya:
Fix should be available in nightly >= 2016-06-08.

sguptatw:
When there are multiple tests failing, repeatedly running gauge --failed is doubling the statistics everytime.

```
Successfully generated html-report to => /Users/sswaroop/work/gauge-test-apps/basic-java/reports/html-report
Specifications:	1 executed	0 passed	1 failed	0 skipped
Scenarios:	32 executed	0 passed	32 failed	0 skipped
```

```
Successfully generated html-report to => /Users/sswaroop/work/gauge-test-apps/basic-java/reports/html-report
Specifications:	1 executed	0 passed	1 failed	0 skipped
Scenarios:	64 executed	0 passed	64 failed	0 skipped
```

apoorvam:
Fix should be available in nightly >= 10.06.2016

sguptatw:
Html report is showing wrong results when the spec is executed with data driven execution.

Atleast it should show the row with the failure.

apoorvam:
Fix should be available in nightly >= 13.06.2016

PR #419, opened 2016-05-27
Don't run default specs if --failed is run with no previous failures

kaustavdm:
- Wrap `runner.SetFlags()` in `runner.Initialize()`.
- `runner.Initialize()` returns returns error if `--failed` flag is set, but no previous failures are found
- `main()` calls `runner.Initialize()` and exits if it returns an error.

This addresses #418. Putting it up for code review suggestions.

### DIFF
--- a/execution/rerun/rerun.go
+++ b/execution/rerun/rerun.go
@@ -19,6 +19,7 @@ package rerun
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


execution/rerun/rerun.go
```diff
@@ -195,5 +191,19 @@ func SetFlags() {
...
 	flag.CommandLine.Parse(meta.FailedItems)
+}
+
+// Initialize sets up rerun and determines whether any failed tests will be executed or not
+func Initialize() error {
+	if !RunFailed {
+		flag.Visit(saveFlagState)
+		return nil
+	}
+	meta := new(failedMetadata)
```
mahendrakariya: There is already a `newFailedMetadata` func. I think it's better to use that.

PR #467, opened 2016-06-23
WIP : generate protocol change for duplicate step error

maquessime:
related to https://github.com/getgauge/gauge-java/pull/59

I've just run genproto.sh after change here : https://github.com/getgauge/gauge-proto/pull/2 

The latter should be merged before merging this one, that's why I've put it as WIP.

Guys, do you think that could cause regression for other runner (ruby, csharp, etc..) if we merge this one?


kashishm:
These changes are updated.

Issue #219, opened 2015-09-28
[Data driven execution] Scenario should be executed once, if it doesnt use dynamic param from data table


apoorvam:
If i have `n` rows in a table at spec level, my scenarios are executed n times. This should be done only if scenario uses any of the params from table. Else, it should be executed only once. 
___________________
Test case
------------
> Note: - In a single spec with multiple scenarios only the scenarios with data table params execute multiple times
- [x] A spec with scenarios without a datatable should execute only once
- [x] A spec with scenarios with a datatable with parameter being used should be executed datarow times.
- [x] A spec with scenarios with, context, teardown a datatable without parameter being used should be executed only once.
- [x] A spec with concept with a datatable with parameter being used should be executed datarow times.
- [x] A spec with scenarios with context having a datatable with parameter being used should be executed datarow times.
- [x] A spec with scenarios with teardown having a datatable with parameter being used should be executed datarow times.
- [ ] This should work if the datatables are executed from the command line with limiting datarow parameters.
- [x] It should work as expected when re-run failed tests.
- [x] It should work as expected with csv file parameter
- [x] It should work as expected when a subset of specs are executed with tags, folders, explicit spec selections.
- [x] It should work as expected when run using the line number.
- [x] The xml-report should have only one entry for the execution
- [ ] The html-report shouldn't show the table rows in the report, or should grey-out the rows.
- [x] The console should not print the rows if it is being executed only once.

sguptatw:
Executing from command line limiting datarow parameters.

**Expected behavior**
The spec having data tables but with scenarios not using the values should behave similar to The spec having no data tables.

**Actual behavior**
The spec with data table does not get executed. The spec having data tables but with scenarios not using the values are getting executed.

PR #600, opened 2017-03-20
#219 table driven execution - Scenarios not using param from data table should be executed only once

Apoorva-GA:


reporter/reporter.go
```diff
@@ -123,14 +124,27 @@ func ListenExecutionEvents() {
 			switch e.Topic {
 			case event.SpecStart:
 				r.SpecStart(e.Item.(*gauge.Specification).Heading.Value)
+				if e.Item.(*gauge.Specification).UsesArgsInContextTeardown(e.Item.(*gauge.Specification).DataTable.Table.Headers...){
```
kashishm: There is no need to store this info. A scenario has a `DataTableRow` field. This scenario field should only be assigned if it is a data table driven scenario and uses a dynamic param. If that field has some value then only print table info.

PR #648, opened 2017-05-17
Concept and spec caching

Apoorva-GA:
- using RWMutex for spec, concept and step cache instead of wait groups.


api/infoGatherer/specDetails.go
```diff
@@ -37,14 +37,28 @@ import (
...
-	stepsCache        map[string][]*gauge.StepValue
+	specsCache        specsCache
+	conceptsCache     conceptCache
+	stepsCache        stepsCache
 	SpecDirs          []string
 }
 
+type conceptCache struct {
+	mutex         sync.RWMutex
+	conceptsCache map[string][]*gauge.Concept
```
sriv: perhaps this field should have a different name? this leads to access such as `conceptCache.conceptsCache` below..

Similar for `stepsCache` and `specsCache`. Perhaps the name of the entity to be used makes sense? i.e. `conceptCache.concepts`?

api/infoGatherer/specDetails.go
```diff
@@ -169,52 +184,61 @@ func (s *SpecInfoGatherer) getParsedConcepts() map[string]*gauge.Concept {
 }
 
 func (s *SpecInfoGatherer) getStepsFromCachedSpecs() map[string][]*gauge.StepValue {
+	s.specsCache.mutex.RLock()
```
sriv: Should probably stick to one paradigm. Elsewhere there is an unlock call using `defer`.

parser/conceptParser_test.go
```diff
@@ -69,7 +71,9 @@ func (s *MySuite) TestDuplicateConceptsinMultipleFile(c *C) {
 	cpt2, _ := filepath.Abs(filepath.Join("testdata", "err", "cpt", "duplicate.cpt"))
 
 	AddConcepts(cpt1, dictionary)
-	errs := AddConcepts(cpt2, dictionary)
+	steps, errs := AddConcepts(cpt2, dictionary)
```
sriv: `steps` => `concepts`  ?

Issue #636, opened 2017-04-26
Not creating the script htmlReport for regeneration


sguptatw:
Running the command 
```
mvn gauge:execute -DspecsDir=specs -DinParallel -Dnodes=16 -Dflags="--strategy=eager"
```
The generated report/htmlReport folder does not contain the htmlReport re-generation script.
[art 2.zip](https://github.com/getgauge/gauge/files/957760/art.2.zip)

**Version**
```
Gauge version: 0.8.4.nightly-2017-04-20

Plugins
-------
html-report (4.0.0.nightly-2017-04-20)
java (0.6.3.nightly-2017-04-21)
xml-report (0.2.0)
```

sriv:
Unable to replicate with below

```
Gauge version: 0.8.4.nightly-2017-04-20

Plugins
-------
csharp (0.10.0)
html-report (4.0.0.nightly-2017-04-25)
java (0.6.2)
ruby (0.4.0)
```

please re-test with latest nightly of html-report.

sguptatw:
Now I am getting
```
runtime: out of memory: cannot allocate 338100224-byte block (2285633536 in use)
fatal error: out of memory

runtime stack:
runtime.throw(0x4c461d, 0xd)
	/usr/local/Cellar/go/1.8/libexec/src/runtime/panic.go:596 +0x76
runtime.largeAlloc(0x1426a000, 0x46500, 0xcce58000)
	/usr/local/Cellar/go/1.8/libexec/src/runtime/malloc.go:797 +0xd5
runtime.mallocgc.func1()
	/usr/local/Cellar/go/1.8/libexec/src/runtime/malloc.go:690 +0x2b
runtime.systemstack(0x10a10a80)
	/usr/local/Cellar/go/1.8/libexec/src/runtime/asm_386.s:337 +0x52
runtime.mstart()
	/usr/local/Cellar/go/1.8/libexec/src/runtime/proc.go:1132

goroutine 1 [running]:
runtime.systemstack_switch()
	/usr/local/Cellar/go/1.8/libexec/src/runtime/asm_386.s:291 fp=0x1163aca4 sp=0x1163aca0
runtime.mallocgc(0x1426a000, 0x0, 0xce49300, 0xcce58000)
	/usr/local/Cellar/go/1.8/libexec/src/runtime/malloc.go:691 +0x691 fp=0x1163acf4 sp=0x1163aca4
[...]

PR #654, opened 2017-05-23
save last run result after suite completes, #636

sriv:
- save the suite result in `.gauge` as `last_run_result`


execution/saveResult.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+		}
+	}()
+}
+
+func writeResult(res *result.SuiteResult) {
+	dotGaugeDir := filepath.Join(config.ProjectRoot, dotGauge)
+	resultFile := filepath.Join(config.ProjectRoot, dotGauge, lastRunResult)
+	if err := os.MkdirAll(dotGaugeDir, common.NewDirectoryPermissions); err != nil {
+		logger.Fatalf("Failed to create directory in %s. Reason: %s", dotGaugeDir, err.Error())
```
kashishm: Use `logger.Error` instead of `Fatalf`.

execution/saveResult.go
```diff
@@ -0,0 +1,71 @@
...
+	dotGauge      = ".gauge"
+	lastRunResult = "last_run_result"
+)
+
+// ListenSuiteEndAndSaveResult listens to execution events and writes the failed scenarios to JSON file
+func ListenSuiteEndAndSaveResult() {
+	ch := make(chan event.ExecutionEvent, 0)
+	event.Register(ch, event.SuiteEnd)
+
+	go func() {
```
kashishm: use `waitGroup` to make sure that the execution of this `goroutine` is complete

Issue #651, opened 2017-05-22
Include additional test result information in ExecutionContext


RabihHallage:
In case of continuous integration or automated deployment pipelines, when running automated test cases the results should be reported into a third party management system/library.
Therefore the need to have all the information related to the test (or step) execution result (e.g. stacktrace in case of failure).
Currently, ExecutionContext includes only the boolean value that indicates whether the execution of a spec, step, or scenario has succeeded or failed.
We Would like to have the mentioned additional information if possible.

Thanks.

PR #657, opened 2017-05-24
Include additional test result information in ExecutionContext #651

RabihHallage:
Include additional test result information in ExecutionContext #651


execution/stepExecutor.go
```diff
@@ -45,7 +45,9 @@ func (e *stepExecutor) executeStep(step *gauge.Step, protoStep *gauge_messages.P
 	if !stepResult.GetFailed() {
 		executeStepMessage := &gauge_messages.Message{MessageType: gauge_messages.Message_ExecuteStep, ExecuteStepRequest: stepRequest}
 		stepExecutionStatus := e.runner.ExecuteAndGetStatus(executeStepMessage)
+		e.currentExecutionInfo.Stacktrace = ""
 		if stepExecutionStatus.GetFailed() {
+			e.currentExecutionInfo.Stacktrace = stepExecutionStatus.GetStackTrace()
```
kashishm: Using `e.currentExecutionInfo.Stacktrace` probably is not a good idea because it does not belong to any entity(`step`, `scenario` or `spec`). We might remove this field in future.

If you want to assign stacktrace for `step` then we can add `stacktrace` field to `e.currentExecutionInfo.CurrentStep` and assign when there is a `step` failure.

For this, you will have to change [gauge-proto](https://github.com/getgauge/gauge-proto/) and add a new field `stacktrace` in [`StepInfo`](https://github.com/getgauge/gauge-proto/blob/master/messages.proto#L112). After that you can use it in `gauge` and `gauge-java`.

RabihHallage: Allright, then i will make the necessary changes!

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "gauge-proto"]
 	path = gauge-proto
-	url = http://github.com/getgauge/gauge-proto
+	url = http://github.com/RabihHallage/gauge-proto
```
kashishm: This should point to `http://github.com/getgauge/gauge-proto`

execution/stepExecutor.go
```diff
@@ -45,7 +45,9 @@ func (e *stepExecutor) executeStep(step *gauge.Step, protoStep *gauge_messages.P
 	if !stepResult.GetFailed() {
 		executeStepMessage := &gauge_messages.Message{MessageType: gauge_messages.Message_ExecuteStep, ExecuteStepRequest: stepRequest}
 		stepExecutionStatus := e.runner.ExecuteAndGetStatus(executeStepMessage)
+		e.currentExecutionInfo.CurrentStep.StackTrace = ""
```
kashishm: This line can be removed. No need to assign the field if there is no failure.

kashishm:
@RabihHallage Thanks for the PR! We really appreciate your contribution.

Issue #658, opened 2017-05-25
Collect statistics for gauge commands


zabil:
**If a user opts-in to send usage information.**

Collect statistics of gauge commands execution to improve or deprecate command line options.
The statistics will be collected via google analytics using [measurement protocol](https://developers.google.com/analytics/devguides/collection/protocol/v1/reference)

This **must** exclude user identifiable information like ip and user specific parameters like spec folders
```
$ gauge myspecfolder
```
Must not record ``myspecfolder``.

or 

```
gauge --dir="../"
```
Must exclude directory information.

sriv:
The default behaviour shall be "opted in", with options for the users to disable these pings. This is so that unattended execution of gauge is not left out.

sriv:
Information captured:

- IP address is **anonymized**
- No Personally Identifyable Information is collected.

- A Client ID (unique UUID)
- Parameters Tracked:
  - Manifest (Language + Plugins)
  - Execution (tagged, parallel, sort, simple-console, verbose flags). No project specific values are captured.
  - Refactor (Rename) (only usage, no parameters)
  - Refactor (Extract To Concept) (only usage, no parameters)
  - Format (only usage, no parameters)
  - List Templates
  - Plugin Install
  - Plugin Update
  - Plugin Install-All
  - Plugin Update-All
  - Plugin Install from Zip
  - Project Init
  - Daemon execution
  - Docs generation (ex spectacle)

---

The raw requests that is sent can be traced by setting `gauge_analytics_log_enabled=true` in `gauge.properties`.

sriv:
should be available in nightly >= 03-jul-2017

sguptatw:
`gauge install` event on a project is not getting logged.

sriv:
> gauge install event without any parameters on a project is not getting logged.

Happened because of fast events, fixed.
> Capture the language, template being used while init

Fixed.
> gauge uninstall <plugin> is not getting logged.

Happened because of fast events, fixed.
> Should a failed request get logged - example failed refactoring, init, add etc?

Telemetry now captures arguments to `init` and `add` subcommands. What is expected to be captured from failed refactoring? Telemetry does not track refactoring text.
> Events which are fast are not getting logged - example format, docs spectacle with one file.

Fixed
> gauge check-updates is not getting logged.

Incorrect command. Try `gauge updates -c` (or `gauge --check-updates` for legacy syntax).

sguptatw:
Please find the test results inline in my above comment.

[...]

Issue #653, opened 2017-05-23
Assign a unique identifier when installing or using Gauge


zabil:
Generate a unique identifier for gauge installation to enable analytics on Gauge.
In the future, if a user **opts-in** to participate, the team can use this for identifying feature usage, support, and adoption. 

Display the identifier when along with the version. 

```
$ gauge -v 
Gauge version: 0.8.4 (xxx-xxxx-xxxxx-xxxx)

Plugins
-------
html-report (4.0.0)
java (0.6.1)
python (0.2.0)
ruby (0.4.0)
spectacle (0.1.2)
```

Where `xxx-xxxx-xxxxx-xxxx` is the identifier

Generate the identifier, if not available, while running Gauge commands.

sguptatw:
- [ ] The ID is generated the first time Gauge is installed/updated
- [ ] Does not change when Gauge is updated later manually
- [ ] Does not change when Gauge is updated later using a package manager
- [ ] Works as expected on all supported OS

PR #665, opened 2017-06-05
introduce install ID, capture anonymous usage stats. #658

sriv:
- generate a unique UUID per install. #653 
- track and send anonymous events to GA, #658

### DIFF
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -110,13 +110,13 @@ func GaugeTemplatesUrl() string {
 // AnalyticsEnabled determines if sending data to analytics is enabled
 func AnalyticsEnabled() bool {
 	e := getFromConfig(analyticsEnabled)
-	return convertToBool(e, checkUpdates, true)
+	return convertToBool(e, analyticsEnabled, true)
 }
 
 // AnalyticsLogEnabled determines if requests to analytics have to be logged
 func AnalyticsLogEnabled() bool {
 	log := getFromConfig(analyticsLoggingEnabled)
-	return convertToBool(log, checkUpdates, false)
+	return convertToBool(log, analyticsLoggingEnabled, false)
 }
 
 // SetProjectRoot sets project root location in ENV.
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -28,14 +28,16 @@ import (
 )
 
 const (
-	gaugeRepositoryUrl      = "gauge_repository_url"
-	gaugeUpdateUrl          = "gauge_update_url"
-	gaugeTemplatesUrl       = "gauge_templates_url"
+	gaugeRepositoryURL      = "gauge_repository_url"
+	gaugeUpdateURL          = "gauge_update_url"
+	gaugeTemplatesURL       = "gauge_templates_url"
 	runnerConnectionTimeout = "runner_connection_timeout"
 	pluginConnectionTimeout = "plugin_connection_timeout"
 	pluginKillTimeOut       = "plugin_kill_timeout"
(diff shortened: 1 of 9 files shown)


skel/gauge.properties
```diff
@@ -20,3 +20,9 @@ runner_request_timeout = 30000
 
 # Allow Gauge and its plugin updates to be notified.
 check_updates = true
+
+# Allow Gauge to collect anonymous usage statistics
+gauge_analytics_enabled = true
```
kashishm: The value should be `false` here. The analytics should be turned off by default.

filter/specItemFilter.go
```diff
@@ -77,25 +76,15 @@ func (filter *ScenarioFilterBasedOnTags) replaceSpecialChar() {
 }
 
 func (filter *ScenarioFilterBasedOnTags) formatAndEvaluateExpression(tagsMap map[string]bool, isTagQualified func(tagsMap map[string]bool, tagName string) bool) (bool, error) {
-	_, tags := filter.getOperatorsAndOperands()
-	expToBeEvaluated := filter.tagExpression
-	sort.Sort(ByLength(tags))
+	tagExpressionParts, tags := filter.parseTagExpression()
```
kashishm: Why are these changes showing up in this PR. These changes are already merged in #671

config/configuration.go
```diff
@@ -121,6 +136,22 @@ func SetProjectRoot(args []string) error {
...
 }
 
+// UniqueID gets the unique installation ID.
+func UniqueID() string {
+	configDir, err := common.GetConfigurationDir()
+	if err != nil {
+		APILog.Warning("Unable to read config dir, %s", err)
+		return ""
+	}
+	idFile := filepath.Join(configDir, ".gauge_id")
```
kashishm: The file is already in `.gauge` folder. So, the filename can just be `.id` or `id`. Thoughts?

config/configuration_test.go
```diff
@@ -46,27 +42,70 @@ func stub4GetFromConfig(propertyName string) string {
...
+	expected := "foo"
+	idFile := filepath.Join("_testData", ".gauge_id")
+	ioutil.WriteFile(idFile, []byte(expected), common.NewFilePermissions)
+
+	s, err := filepath.Abs("_testData")
+	if err != nil {
+		t.Error(err)
+	}
+
+	os.Setenv("GAUGE_ROOT", s)
```
kashishm: We should be using `GAUGE_HOME`. `GAUGE_ROOT` is just for backward compatibility.

sriv: we use `common.GetConfigurationDir` which returns `gauge_root` or `$home/.gauge/config`

track/track.go
```diff
@@ -0,0 +1,204 @@
...
+
+func Execution(parallel, tagged, sorted, simpleConsole, verbose bool, parallelExecutionStrategy string) {
+	action := "serial"
+	if parallel {
+		action = "parallel"
+	}
+
+	label := ""
+	if tagged {
+		label = "tagged,"
```
kashishm: Why is `,` required at the end. If I am execution via tags(no other flag) then the label would be `tagged,`. Is that intentional? May be we can use array and join it at the end with `,`

config/configuration_test.go
kashishm: Ok. I think we should change that also to use `GAUGE_HOME`. But this is out of scope for this PR

track/track.go
sriv: fair point. sorted in the next commit.

PR #695, opened 2017-06-30
Spec format adding newlines before and after tags

Apoorva-GA:


formatter/formatSpecification.go
```diff
@@ -29,27 +29,47 @@ type formatter struct {
 	buffer bytes.Buffer
 }
 
+var nextItem gauge.Item
```
sriv: why not have `nextItem` in the `formatter` type?

formatter/formatSpecification.go
```diff
@@ -29,27 +29,47 @@ type formatter struct {
...
+
+func (formatter *formatter) SetNext(item gauge.Item) {
+	nextItem = item
+}
+
+func (formatter *formatter) GetNext() (item gauge.Item) {
+	return nextItem
+}
+
```
sriv: `GetNext` is not idiomatic go code. Ref : https://golang.org/doc/effective_go.html#Getters

In any case, if you have `nextItem` in the struct, you need not have a `Getter`.

formatter/formatSpecification.go
```diff
@@ -29,27 +29,47 @@ type formatter struct {
...
+	}
 }
 
-func (formatter *formatter) SpecTags(tags *gauge.Tags) {
+func (formatter *formatter) Tags(tags *gauge.Tags) {
+	if !strings.HasSuffix(formatter.buffer.String(), "\n\n") {
+		formatter.buffer.WriteString("\n")
+	}
 	formatter.buffer.WriteString(FormatTags(tags))
+	if formatter.GetNext().Kind() != gauge.CommentKind || strings.TrimSpace(nextItem.(*gauge.Comment).Value) != "" {
```
sriv: mixed use of `formatter.GetNext` and `nextItem`. Hence the proposal to have `nextItem` in `formatter`.

formatter/formatSpecification.go
```diff
@@ -29,27 +29,47 @@ type formatter struct {
...
-	formatter.buffer.WriteString(FormatExternalDataTable(extDataTable))
-}
-
-func (formatter *formatter) ContextStep(step *gauge.Step) {
-	formatter.Step(step)
+func (formatter *formatter) DataTable(dataTable *gauge.DataTable) {
+	if !dataTable.IsExternal {
+		formatter.Table(&(dataTable.Table))
+	} else {
+		formatter.buffer.WriteString(FormatExternalDataTable(dataTable))
```
sriv: If `FormatExternalDataTable` is not used outside the package, it should not be exported.

gauge/ItemQueue.go
```diff
@@ -0,0 +1,28 @@
+package gauge
```
sriv: Please include the license header

Issue #685, opened 2017-06-29
Statistics inconsistent when specs are repeated


sguptatw:
_From @sguptatw on June 29, 2017 7:47_

**Expected behavior**
The specifications executed count should be the number of specifications executed. i.e If any specs are repeated, then it should be listed as many times.

**Actual behavior**
The scenario count reflects the repeated execution, the specifications dont.

**Steps to replicate**
* Run specs from the command line 
```
gauge run specs -d="ttt"
# Specification Heading
  ## Vowel counts in single word	 ✔ ✔
  ## Vowel counts in multiple word	 ✔ ✔

Successfully generated html-report to => /Users/sswaroop/work/test/ttt/reports/html-report
Specifications:	1 executed	1 passed	0 failed	0 skipped
Scenarios:	2 executed	2 passed	0 failed	0 skipped
```
* Now run the following command
```
gauge run specs/example.spec specs -d="ttt"
# Specification Heading
  ## Vowel counts in single word	 ✔ ✔
  ## Vowel counts in multiple word	 ✔ ✔
  ## Vowel counts in single word	 ✔ ✔
  ## Vowel counts in multiple word	 ✔ ✔

Successfully generated html-report to => /Users/sswaroop/work/test/ttt/reports/html-report
Specifications:	1 executed	1 passed	0 failed	0 skipped
Scenarios:	4 executed	4 passed	0 failed	0 skipped
```
The specification count is 1 and the scenario count is 4.

**Version**
```
Gauge version: 0.9.0.nightly-2017-06-28

Plugins
-------
csharp (0.10.1)
html-report (4.0.1)
java (0.6.4)
ruby (0.4.1)
spectacle (0.1.2)
```

_Copied from original issue: getgauge/html-report#151_

nehashri:
The approach to this has changed. If we specify duplicate specifications, the specification will be run only once(only the first specified specification is run). The same applies for scenarios also.

Apoorva-GA:
Available in nightly > 07/07/17

PR #701, opened 2017-07-03
Not appending duplicate scenario to spec. #685

nehashri:


nehashri:
If there are indexed scenarios specified, we can't just perform de-dup based on spec file name. Hence, we are removing duplicate spec files given by the user at the parser level and we are appending the scenarios explicitly. This change only handles the duplicates on the scenarios if given.

nehashri:
Closing this pull request. Refer pull request https://github.com/getgauge/gauge/pull/707

PR #734, opened 2017-07-25
add build support for analytics export

sriv:
generates monthly export of Google Analytics data. 

published at https://insights.getgauge.io

### DIFF
--- a/build/analytics/dump.go
+++ b/build/analytics/dump.go
@@ -1,0 +1,403 @@
+// Copyright 2015 ThoughtWorks, Inc.
+
+// This file is part of Gauge.
+
+// Gauge is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


build/analytics/dump.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+func main() {
+	jwtc := jwt.Config{
+		Email:      gaServiceAcctEmail,
+		PrivateKey: []byte(gaServiceAcctPEM),
+		Scopes:     []string{analytics.AnalyticsReadonlyScope},
+		TokenURL:   tokenurl,
+	}
+	clt := jwtc.Client(oauth2.NoContext)
```
nehashri: `oauth2.NoContext` seems to be deprecated. We should use an alternative.

PR #735, opened 2017-07-26
improvements on LSP support for code completion

nehashri:
- completion shows only if the cursor is in between a step
- completion doesn't show up if cursor is in between parameters 
- duplicate steps are filtered
- completion replaces the text (after *) entered by the user.

### DIFF
--- a/api/lang/completion_test.go
+++ b/api/lang/completion_test.go
@@ -409,3 +409,65 @@ func TestGetFilterTextWithLesserNumberOfStepArgsGiven(t *testing.T) {
 		t.Errorf("The parameters are not replaced correctly")
 	}
 }
+
+var testEditPosition = []struct {
+	input     string
+	cursorPos lsp.Position
+	wantStart lsp.Position
+	wantEnd   lsp.Position
+}{
+	{
+		input:     "*",
+		cursorPos: lsp.Position{Line: 0, Character: 1},
+		wantStart: lsp.Position{Line: 0, Character: 1},
+		wantEnd:   lsp.Position{Line: 0, Character: 1},
+	},
+	{
+		input:     "* ",
+		cursorPos: lsp.Position{Line: 0, Character: 1},
+		wantStart: lsp.Position{Line: 0, Character: 1},
+		wantEnd:   lsp.Position{Line: 0, Character: 2},
+	},
+	{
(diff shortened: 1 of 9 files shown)


api/lang/completion_test.go
```diff
@@ -98,28 +101,204 @@ func TestCompletion(t *testing.T) {
...
-	b, _ := json.Marshal(lsp.TextDocumentPositionParams{})
+
+	b, _ := json.Marshal(lsp.TextDocumentPositionParams{TextDocument: lsp.TextDocumentIdentifier{URI: "uri"}, Position: position})
 	p := json.RawMessage(b)
 
 	got, err := completion(&jsonrpc2.Request{Params: &p})
 
 	if err != nil {
-		t.Errorf("Expected error == nil in Completion, got %s", err.Error())
+		t.Fatalf("Expected error == nil in Completion, got %s", err.Error())
```
sriv: Why `Fatalf` instead of `Errorf` here?

api/lang/completion_test.go
```diff
@@ -98,28 +101,204 @@ func TestCompletion(t *testing.T) {
...
+	}
+	provider = &dummyCompletionProvider{}
+
+	b, _ := json.Marshal(lsp.TextDocumentPositionParams{TextDocument: lsp.TextDocumentIdentifier{URI: "uri"}, Position: position})
+	p := json.RawMessage(b)
+
+	got, err := completion(&jsonrpc2.Request{Params: &p})
+
+	if err != nil {
+		t.Fatalf("Expected error == nil in Completion, got %s", err.Error())
```
sriv: `Fatalf` instead of `Errorf`?

api/lang/completion_test.go
nehashri: Changed this mainly because Errorf does not fail the test now and exit. The next check present depends on the `err` not being nil. Fatalf does not move forward and fails the test completely at that point. Hence the change.

api/lang/completion_test.go
nehashri: Changed this mainly because Errorf does not fail the test now and exit. The next check present depends on the `err` not being nil. Fatalf does not move forward and fails the test completely at that point. Hence the change.

Issue #736, opened 2017-07-26
Allow telemetry to be turned on/off via env var


sriv:
## Actual Behaviour

Telemetry in gauge can be turned on or off using 

`gauge telemetry on` or `gauge telemetry off`

The status of telemetry can be queried via 

    gauge telemetry.

## Expected Behaviour

Introduce an environment variable `GAUGE_TELEMETRY_ENABLED` (true|false) which will get priority over the value of `gauge_telemetry_enabled` property in `gauge.properties`.

When GAUGE_TELEMETRY_ENABLED env var is set:

- `gauge telemetry on` and `gauge telemetry off` will continue to modify the value in `gauge.properties`
it will print a warning stating "Telemetry status is controlled via GAUGE_TELEMETRY_ENABLED environment variable"

- `gauge telemetry` will print the effective value. it will print an info stating "Telemetry status is controlled via GAUGE_TELEMETRY_ENABLED environment variable"
-

sguptatw:
When GAUGE_TELEMETRY_ENABLED is off, the `gauge telemetry on` should indicate that the env value has to be on for telemetry to be on.

PR #737, opened 2017-07-26
allow telemetry to be overridden via env var, #736

sriv:
- GAUGE_TELEMETRY_ENABLED env var will override `gauge telemetry` and control the toggle.
- print message when GAUGE_TELEMETRY_ENABLED is set.

### DIFF
--- a/cmd/telemetry.go
+++ b/cmd/telemetry.go
@@ -19,6 +19,8 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"strings"
 
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


cmd/telemetry.go
```diff
@@ -27,6 +29,24 @@ import (
...
 )
 
+const gaugeTelemetryEnabled = "GAUGE_TELEMETRY_ENABLED"
+
+var telemetryEnv = os.Getenv(gaugeTelemetryEnabled)
+
+func telemetryEnabled() bool {
+	if telemetryEnv == "" {
+		return config.TelemetryEnabled()
+	}
```
BugDiver: Maybe this check can be removed. If `telemetryEnv` is empty `strconv.ParseBool` will return err which is handled in next check.

sriv: This is because we print the message only if `GAUGE_TELEMETRY_ENABLED` is set. If we remove this check, there needs to be an additional check in Line-44.

BugDiver: if `telemetryEnv` is empty then `strconv.ParseBool` will fail and return `config.TelemetryEnabled()` which is what the extra check is returning. It won't even reach to Line 44.

Issue #745, opened 2017-07-28
Add option to disable telemetry via mac and linux installer


Apoorva-GA:
### Expected behavior
Tell us what should happen

### Actual behavior
Tell us what happens instead

### Steps to reproduce
1.
2.
3.

### Gauge version
```
Run gauge -v on your system and paste the results here.
```

sguptatw:
Test cases
I have to get the message
- [ ] for choco installation
- [x] check box for mac pkg installer
- [x] check box for windows installer
- [ ] Message while installing using brew
- [x] Installing from zip
- [x] deb

sriv:
chocolatey installer has the message printed now.

sguptatw:
brew and choco messages to be tested after release

sguptatw:
choco installation message is available. However brew install does not have the message.

PR #746, opened 2017-07-28
Option to turn off telemetry via mac and linux installers #745

Apoorva-GA:


build/install/install.sh
```diff
@@ -221,6 +221,7 @@ install_gauge_interactively() {
     copy_gauge_binaries_interactively
     add_gauge_in_path
     echo -e "Gauge core successfully installed.\n"
+    echo -e "We are constantly looking to make Gauge better, and report usage statistics anonymously over time. If you do not want to participate please read instructions https://manpage.getgauge.io/gauge_telemetry_off.html on how to turn it off.\n"
 }
```
BugDiver: Multiple usage. Can be extracted.

Issue #756, opened 2017-08-03
Formatting tags defined in multiple lines.


Apoorva-GA:
Tags defined in multiple lines should be formatted.
Given tags : 
```
tags: def, 
abc, asd, 
xyz
```

### Expected behavior
```
tags: def, 
      abc, asd, 
      xyz
```
### Actual behavior
```
tags: def, abc, asd, xyz
```

Apoorva-GA:
Should be available in nightly >= 9/8/17

sguptatw:
The preview in the browser still shows all the tags in a single line.

Vishnu1067:
@Apoorva-GA I tried to use tags in multiple lines and second line considered as description

```Tags: smoke-newdash-sg, smoke-newdash-my, smoke-newdash-id, smoke-newdash-dd, smoke-newdash-dde, smoke-newdash-mobile-sg,
      smoke-newdash-mobile-my, smoke-newdash-mobile-id, smoke-newdash-mobile-dd, smoke-newdash-mobile-dde```

sguptatw:
Gauge does supports tags in multiple lines. Executing gauge run specs --tags="smoke-newdash-mobile-my" from the command line executes the corresponding specs.

However, there are changes required on IDE plugins to highlight the tags in multiple lines. These  are being tracked separately.
Please follow issues getgauge/Intellij-Plugin#285 and getgauge/gauge-visualstudio#194 for updates.

PR #757, opened 2017-08-03
Formatting tags defined in multiple lines #756

Apoorva-GA:


gauge/specification.go
```diff
@@ -310,9 +310,19 @@ func (t *TearDown) Kind() TokenKind {
...
 type Tags struct {
-	Values []string
+	Values [][]string
 }
 
+func (tags *Tags) Add(values []string) {
+	tags.Values = append(tags.Values, values)
+}
+
+func ToArray(values [][]string) (val []string) {
```
sriv: `gauge.ToArray` seems to be called only for `tags.Values` for flatenning a 2d array into one. can we instead have the method on tags struct and avoid the scatter of `gauge.ToArray` everywhere?

PR #766, opened 2017-08-10
Generating template for unimplemented  step.

Apoorva-GA:


cmd/validate.go
```diff
@@ -24,23 +24,31 @@ import (
...
-}
+var (
+	validateCmd = &cobra.Command{
+		Use:     "validate [flags] [args]",
+		Short:   "Check for validation and parse errors",
+		Long:    `Check for validation and parse errors.`,
+		Example: "  gauge validate specs/",
+		Run: func(cmd *cobra.Command, args []string) {
+			setGlobalFlags()
+			if hideQuickFix {
```
sriv: can be just `validation.HideQuickFix=hideQuickFix`

cmd/validate.go
```diff
@@ -24,23 +24,31 @@ import (
...
+				validation.HideQuickFix = hideQuickFix
+			}
+			if err := isValidGaugeProject(args); err != nil {
+				logger.Fatalf(err.Error())
+			}
+			initPackageFlags()
+			track.Validation()
```
sriv: Add `hide-quick-fix` to tracking as a label.

validation/validate.go
```diff
@@ -66,7 +68,7 @@ type SpecValidationError struct {
 }
 
 func (s StepValidationError) Error() string {
-	return fmt.Sprintf("%s:%d %s => '%s'", s.fileName, s.step.LineNo, s.message, s.step.GetLineText())
+	return fmt.Sprintf("%s:%d %s => '%s'%s", s.fileName, s.step.LineNo, s.message, s.step.GetLineText(), s.quickFix)
```
sriv: Check with @zabil if the quick fixes need to be inline with the error, as against a summary at the end of the run

cmd/validate.go
```diff
@@ -24,23 +24,31 @@ import (
...
+			validation.Validate(args)
+		},
+		DisableAutoGenTag: true,
+	}
+	hideQuickFix bool
+)
 
 func init() {
 	GaugeCmd.AddCommand(validateCmd)
+	validateCmd.Flags().BoolVarP(&hideQuickFix, "hide-quick-fix", "", false, "Gives default step implementation for un implemented steps")
```
sriv: `QuickFix` is a Intellij specific naming. Something generic like `Suggestion` makes sense.

gauge.go
```diff
@@ -159,7 +159,7 @@ func main() {
 			track.Format()
 			formatter.FormatSpecFilesIn(*specFilesToFormat)
 		} else if *validate {
-			track.Validation()
+			track.Validation(*validate)
```
sriv: Why pass `validate` flag? This will be `true` always. Instead it should be `false` since the `hide-suggestion` flag does not exist in legacy usage.

validation/validate.go
```diff
@@ -66,7 +68,7 @@ type SpecValidationError struct {
 }
 
 func (s StepValidationError) Error() string {
-	return fmt.Sprintf("%s:%d %s => '%s'", s.fileName, s.step.LineNo, s.message, s.step.GetLineText())
+	return fmt.Sprintf("%s:%d %s => '%s'%s", s.fileName, s.step.LineNo, s.message, s.step.GetLineText(), s.suggestion)
```
sriv: @zabil had some ideas on the format of the error to be printed. Perhaps it could be included here.

Apoorva-GA: Talked to Zabil. Decide to keep it inline for now. Will implement his ideas separately as it needs some more thinking.

cmd/validate.go
```diff
@@ -24,23 +24,29 @@ import (
...
+			validation.Validate(args)
+		},
+		DisableAutoGenTag: true,
+	}
+	hideSuggestion bool
+)
 
 func init() {
 	GaugeCmd.AddCommand(validateCmd)
+	validateCmd.Flags().BoolVarP(&hideSuggestion, "hide-suggestion", "", false, "Gives default step implementation for un implemented steps")
```
sriv: "unimplemented" is a single word. Also lets not call it "default" step implementation. How about message like this:

"Prints a step implementation stub for every unimplemented step"

Issue #772, opened 2017-08-16
Multiple errors with same cause do not have suggestions


sriv:
### Expected behavior
When more than one specs have validation errors, all of them should be listed when running `gauge validate specs` or `gauge run specs`. If suggestions are available, they should be printed for all errors that can have suggestions.

### Actual behavior
If more than one specs have validation errors, all of them are listed when running `gauge validate specs` or `gauge run specs`. However, not all errors contain suggestions.

### Steps to reproduce
1. Download the project [multi_errors_no_suggestion.zip](https://github.com/getgauge/gauge/files/1227283/multi_errors_no_suggestion.zip)

2. Run `gauge run specs`
3. Observe the output:

```
[ValidationError] /tmp/foo/specs/example.spec:18 Step implementation not found => 'The word "gauge" has "3" vowels.'

Quick Fix :
@Step("The word <gauge> has <3> vowels.")
public void implementation(Object arg0, Object arg1){
}

[ValidationError] /tmp/foo/specs/example1.spec:18 Step implementation not found => 'The word "gauge" has "3" vowels.'
[ValidationError] /tmp/foo/specs/example2.spec:18 Step implementation not found => 'The word "gauge" has "3" vowels.'
[ValidationError] /tmp/foo/specs/example3.spec:18 Step implementation not found => 'The word "gauge" has "3" vowels.'
# Specification Heading
  ## Vowel counts in multiple word	 ✔ ✔

# Specification Heading
  ## Vowel counts in multiple word	 ✔ ✔

# Specification Heading
  ## Vowel counts in multiple word	 ✔ ✔

# Specification Heading
  ## Vowel counts in multiple word	 ✔ ✔

Successfully generated html-report to => /tmp/foo/reports/html-report
Specifications:	4 executed	4 passed	0 failed	0 skipped
Scenarios:	4 executed	4 passed	0 failed	4 skipped

Total time taken: 48ms
```

### Gauge version
```
Gauge version: 0.9.2 [built from source]

Plugins
-------
csharp (0.10.1)
html-report (4.0.1)
java (0.6.5)
js (2.0.0)
ruby (0.4.1)
xml-report (0.2.0)
```

Apoorva-GA:
If there is a unimplemented step in multiple files. How should the suggestion be shown? Should it be shown for each unimplemented step or consolidate all validation error for that particular step and show suggestion at the end?

Something like this : 
CASE : 1
```
[...]

PR #775, opened 2017-08-18
Showing validation errors first and then the suggestions #772

Apoorva-GA:


validation/validation_test.go
```diff
@@ -145,11 +145,13 @@ func (s *MySuite) TestValidateStep(c *C) {
...
 	c.Assert(valErr, Not(Equals), nil)
-	c.Assert(valErr.Error(), Equals, "foo.spec:3 Step implementation not found => 'my step'"+"\n"+
-		"Suggestion : "+"\n"+
-		"@Step(\"my step\")"+"\n"+
-		"public void implementation(){"+"\n\t"+
-		"// your code here...\n}")
+	c.Assert(valErr.Error(), Equals, "foo.spec:3 Step implementation not found => 'my step'")
+	c.Assert(valErr.(StepValidationError).Suggestion(), Equals, "One or more errors were due to : Step implementation not found => 'my step'\t\n"+
+		"Suggestion : \n"+
+		"@Step(\"my step\")\n"+
```
sriv: We should have the suggestion indented so that it looks something like this:

```
Suggestion :
    @Step("my step")
    public void implementation(){
        // your code here...
    }
```

kashishm: This will not work for languages having indentation-based blocks like `python`.

validation/validate.go
```diff
@@ -68,7 +68,11 @@ type SpecValidationError struct {
 }
 
 func (s StepValidationError) Error() string {
-	return fmt.Sprintf("%s:%d %s => '%s'%s", s.fileName, s.step.LineNo, s.message, s.step.GetLineText(), s.suggestion)
+	return fmt.Sprintf("%s:%d %s => '%s'", s.fileName, s.step.LineNo, s.message, s.step.GetLineText())
+}
+
+func (s StepValidationError) Suggestion() string {
+	return fmt.Sprintf("%s : %s => '%s'\t%s", "One or more errors were due to", "Step implementation not found", s.step.GetLineText(), s.suggestion)
```
kashishm: Why the suggestion message `Step implementation not found` is hardcoded? The error type is `StepValidationError` that means it can be used for both`step implementation not found` and `duplicate step implementations found` errors.

sriv: Agreed, It should show the `StepValidationError.message`.

Also it should check the `errorType` so that suggestion is valid for the given errortype.

validation/validation_test.go
sriv: @kashishm - fair enough. I think that the language runner should be able to set the suggestion. So `python` may choose to not indent, whereas other's might. From Gauge's point of view, it should print whatever the runner suggests.

PR #779, opened 2017-08-22
Displaying commit SHA with gauge version

Apoorva-GA:


cmd/version.go
```diff
@@ -59,10 +59,11 @@ func PrintJSONVersion() {
 		Version string `json:"version"`
 	}
 	type versionJSON struct {
-		Version string        `json:"version"`
-		Plugins []*pluginJSON `json:"plugins"`
+		Version    string        `json:"version"`
+		CommitHash string        `json:"commitHash"`
```
sriv: Worth being consistent in naming. Either use `hash` or `SHA`.

PR #784, opened 2017-08-26
Update README

sghill:
Noticed a couple small things reading through the docs I thought I could contribute:
* Fix 404 on Contributing link
* Fix Stack Overflow markdown link syntax


zabil:
Oops and thanks!

Issue #676, opened 2017-06-15
Change gauge command structure to have subcommands with flags


kashishm:
## Old Usage

```
Gauge version 0.8.4
Copyright 2017 ThoughtWorks, Inc.

Usage:
	gauge specs/
	gauge specs/spec_name.spec

Options:
  --add-plugin=""                  Adds the specified non-language plugin to the current project
  --api-port=""                    Specifies the api port to be used. Eg: gauge --daemonize --api-port 7777
  --check-updates=false            Checks for Gauge and plugins updates. Eg: gauge --check-updates
  --daemonize=false                Run as a daemon
  --dir="."                        Set the working directory for the current command, accepts a path relative to current directory.
  --docs=""                        Generate documenation using specified plugin. Eg: gauge --docs <plugin name> specs/
  --env="default"                  Specifies the environment. If not specified, default will be used
  --failed=false                   Run only the scenarios failed in previous run. Eg: gauge --failed
  --file, -f=""                    Installs the plugin from zip file. This is used with --install. Eg: gauge --install java -f ZIP_FILE
  --format=""                      Formats the specified spec files
  -g, --group=-1                   Specify which group of specification to execute based on -n flag
  --init=""                        Initializes project structure in the current directory. Eg: gauge --init java
  --install=""                     Downloads and installs a plugin. Eg: gauge --install java
  --install-all=false              Installs all the plugins specified in project manifest, if not installed. Eg: gauge --install-all
  --list-templates=false           Lists all the Gauge templates available. Eg: gauge --list-templates
  --log-level=""                   Set level of logging to debug, info, warning, error or critical
  --machine-readable=false         Used with `--version` to produce JSON output of currently installed Gauge and plugin versions. e.g: gauge --version --machine-readable
  -n=8                             Specify number of parallel execution streams
  --order, -o=""                   Set the specs execution order. Possible options are: `random`, `sort`. Ex: gauge --order="random"
  --parallel, -p=false             Execute specs in parallel
  --plugin-args=""                 Specified additional arguments to the plugin. This is used together with --add-plugin
  --plugin-version=""              Version of plugin to be installed. This is used with --install or --uninstall flag.
  --refactor=""                    Refactor steps. Eg: gauge --refactor <old step> <new step> [[spec directories]]
  --simple-console=false           Removes colouring and simplifies from the console output
  --sort, -s=false                 Run specs in Alphabetical Order. Eg: gauge -s specs
[...]

PR #786, opened 2017-08-26
Update issue template version command

sghill:
On Gauge 0.9.1, the command in the Github issue template to get the version prints this deprecation warning:

```
$ gauge -v
[DEPRECATED] This usage will be removed soon. Run `gauge help --legacy` for more info.
Gauge version: 0.9.1
```

This replaces the deprecated `-v` flag with preferred `version` subcommand.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -21,5 +21,5 @@ Tell us what happens instead
 
 ### Gauge version
 ```
-Run gauge -v on your system and paste the results here.
+Run gauge version on your system and paste the results here.
 ```


sriv:
@sghill - thanks for looking into this. The reason this is left as it is, is because users using `gauge` < `0.9.0` will not see `version` subcommand.

We plan to remove the deprecated usages soon, post which we can update the issue template.

sghill:
@sriv thanks for the followup!

As an alternative, have you considered leaving the `-v` and `--version` options undeprecated alongside the `version` subcommand? `git` does this, so there is some precedent.

sriv:
@sghill - we did not specifically think and discard the `--version` or `-v` flag, however we did revisit the older gauge (<=0.8.5) options and flags, and we realized that `commands` and `options` were getting mixed up and more importantly, all commands and flags were a flattenned list, thus making it very difficult for users to discover the right combination.

But I guess the `-v` and `--version` flags could be left alone, they seem to be more commonly used across various tools. Thanks for the suggestion.

Please see #676 for the change details.

sriv:
@sghill - if it is alright with you, may I close this PR? If you could raise an issue to address the non-deprecation of `-v` and `--version` flags, that would be awesome. Cheers!

sghill:
@sriv closing this is good with me 👍. I opened a new issue to undeprecate (#790). Thanks!

PR #796, opened 2017-09-04
Showing step line no in JSON logger #getgauge/intellij-plugin120

BugDiver:
listening StepEnd even and caching steps corresponding to current
scenario.

### DIFF
--- a/reporter/jsonConsole.go
+++ b/reporter/jsonConsole.go
@@ -54,6 +54,11 @@ type jsonConsole struct {
 	stream     int
 }
 
+type stepInfo struct {
+	step      *gauge.Step
+	protoStep *gm.ProtoStep
+}
+
 type executionEvent struct {
 	EventType eventType        `json:"type"`
 	ID        string           `json:"id,omitempty"`
@@ -84,8 +89,11 @@ type executionError struct {
 	Text       string `json:"text"`
 	Filename   string `json:"filename"`
 	Message    string `json:"message"`
+	LineNo     string `json:"lineNo"`
 	StackTrace string `json:"stackTrace"`
 }
+
+var stepCache = make(map[*gm.ScenarioInfo][]*stepInfo)
 
 func newJSONConsole(out io.Writer, isParallel bool, stream int) *jsonConsole {
 	return &jsonConsole{Mutex: &sync.Mutex{}, writer: out, isParallel: isParallel, stream: stream}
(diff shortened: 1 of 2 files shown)


gaugeci:

reporter/jsonConsole.go
```diff
@@ -84,9 +89,12 @@ type executionError struct {
 	Text       string `json:"text"`
 	Filename   string `json:"filename"`
 	Message    string `json:"message"`
+	LineNo     string `json:"lineNo"`
 	StackTrace string `json:"stackTrace"`
 }
 
+var stepCache = make(map[*gm.ScenarioInfo][]*stepInfo)
```
kashishm: This cache can be added to JSONConsole instead of a global var.

BugDiver: In that case, the private methods who are using the cache has to be bound with JSONConsole

kashishm: Or it can be passed as param.

reporter/jsonConsole.go
```diff
@@ -200,6 +207,10 @@ func (c *jsonConsole) StepStart(stepText string) {
 }
 
 func (c *jsonConsole) StepEnd(step gauge.Step, res result.Result, execInfo gm.ExecutionInfo) {
+	si := new(stepInfo)
```
kashishm: Simplify this to `&stepInfo{step: &step, protoStep: res.(*result.StepResult).Item().(*gm.ProtoStep)}`

reporter/jsonConsole.go
```diff
@@ -183,7 +190,7 @@ func (c *jsonConsole) ScenarioEnd(scenario *gauge.Scenario, res result.Result, i
 		Res: &executionResult{
 			Status:            getScenarioStatus(res.(*result.ScenarioResult)),
 			Time:              res.ExecTime(),
-			Errors:            getErrors(getAllStepsFromScenario(res.(*result.ScenarioResult).ProtoScenario), i.CurrentSpec.FileName),
+			Errors:            getErrors(c, getAllStepsFromScenario(res.(*result.ScenarioResult).ProtoScenario), i.CurrentSpec.FileName, i),
```
kashishm: Is there any specific reason to pass the JSONConsole instead of passing the cache?

BugDiver: No. We can pass `stepCache` instead of `JSONCosole`.

Issue #806, opened 2017-09-13
Parse error: Duplicate concept definition found should give line number and filename of the duplicate concepts


Apoorva-GA:
### Expected behaviour
```
[ParseError] Duplicate concept definition found => 'concept' => at
	/Users/apoorvag/test/concept.cpt:1
	/Users/apoorvag/test/concept.cpt:4
```
### Actual behavior
```
[ParseError] /Users/apoorvag/test/concept.cpt:4 Duplicate concept definition found => 'concept'
```
### Steps to reproduce
1.concept.cpt
```
# concept
* abcd

# concept
* xyz
```
2. Run 'gauge validate specs'

Apoorva-GA:
Should be available in gauge nightly >= 18/09/17

sguptatw:
Test cases
- [ ] Duplicate concept definitions in a single file should be listed.
- [ ] Duplicate concept definitions across files should be listed.
- [ ] Duplicate concept definition list should be independent of the usage in a nested concept
**Expected behavior**
Duplicate concept definition list should be independent of the usage in a nested concept
**Actual behavior**
Duplicate Concept Definition nested in a concept should lists the caller concept as well

**Steps to replicate**
* Create a concept
```
# one
* step2
* step0 "HEAD"

# abc <HEAD>
* step1
* one

# one
* step2
* step0 "HEAD"
```
* Create a spec
```
Specification Heading
=====================
Scenario 1
----------
* step1
* one
```
Now run the specs we get
```
[ParseError] Duplicate concept definition found => 'one' => at
	.../specs/abc.cpt:1
	.../specs/abc.cpt:5
        .../specs/abc.cpt:9
```
However the list mentions concept `abc <HEAD>` as well

**Version**
```
Gauge version: 0.9.3.nightly-2017-09-18
Commit Hash: 952349f

Plugins
-------
csharp (0.10.2.nightly-2017-07-25)
html-report (4.0.2.nightly-2017-08-07)
java (0.6.6.nightly-2017-09-14)
```

PR #807, opened 2017-09-14
ParseError: Duplicate concept definition found should give all the line numbers and filenames of the duplicate concepts #806

Apoorva-GA:


parser/specparser.go
```diff
@@ -803,7 +803,11 @@ type ParseError struct {
 }
 
 func (se ParseError) Error() string {
-	return fmt.Sprintf("%s:%d %s => '%s'", se.FileName, se.LineNo, se.Message, se.LineText)
+	if se.LineNo == 0 && se.FileName == "" {
+		return fmt.Sprintf("%s", se.Message)
+	} else {
```
BugDiver: The `else`  is not needed, already returning from `if` block.

parser/conceptParser.go
```diff
@@ -240,22 +240,44 @@ func CreateConceptsDictionary() (*gauge.ConceptDictionary, *ParseResult) {
...
+		for _, conceptStep := range concepts {
+			if _, exists := conceptDictionary.ConceptsMap[conceptStep.Value]; exists {
+				buffer.WriteString(fmt.Sprintf("\n\t%s:%d", conceptFile, conceptStep.LineNo))
+				message = fmt.Sprintf("Duplicate concept definition found => '%s' => at%s", conceptStep.LineText, buffer.String())
+			} else {
+				conceptDictionary.ConceptsMap[conceptStep.Value] = &gauge.Concept{conceptStep, conceptFile}
+				buffer.WriteString(fmt.Sprintf("\n\t%s:%d", conceptFile, conceptStep.LineNo))
+			}
+			conceptDictionary.ReplaceNestedConceptSteps(conceptStep)
 		}
```
BugDiver: This can be extracted as a method.

Issue #813, opened 2017-09-18
Formatting for LSP : Specification


BugDiver:
Gauge LSP server should support formatting specs.
When an LSP client asks gauge to format a spec file, gauge should format the spec.
This story is part of [LSP support](https://github.com/getgauge/gauge/issues/717) for gauge.

## Assumptions
-  Gauge format feature is working as expected

## Test cases
- [x] Should invoke gauge format if there are no parse errors

**If formatting fails, the reason must be indicated**
- [x]  Table doesn’t belong to any step
- [x] Table location not specified
- [x] Scenario should be defined after the spec heading - This should consider 3 `-` and more as a step. This parse error appears even if there is 1 hyphen
- [x] Could not resolve table from file - give a wrong path in the table parameter <table:something> or <table:>
- [x] Update the vs-code readme with every new feature

sguptatw:
Assumptions
-  Gauge format feature is working as expected

Test cases
- [x] Should invoke gauge format if there are no parse errors
- [x] If formatting fails, the reason must be indicated

> Update the vs-code readme with every new feature

PR #819, opened 2017-09-22
Added spec formatting support through lsp #813

BugDiver:


api/lang/format.go
```diff
@@ -0,0 +1,30 @@
...
+	logger.APILog.Debugf("LangServer: request received : Type: Format Document URI: %s", params.TextDocument.URI)
+	file := convertURItoFilePath(params.TextDocument.URI)
+	if common.FileExists(file) && util.IsValidSpecExtension(file) {
+		res := formatter.FormatSpecFiles(file)
+		if len(res[0].ParseErrors) > 0 {
+			logger.APILog.Debugf("LangServer: Failed to format file : %s", params.TextDocument.URI)
+			return nil, errors.New(res[0].ParseErrors[0].Message)
+		}
+	}
+	return nil, nil
```
kashishm: I think Gauge need not do the formatting, it just has to return the textEdits describing the modifications.

That is what is described in the [protocol](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#document-formatting-request) for Document Formatting Request's response.

> result: TextEdit[] describing the modification to the document to be formatted.

api/lang/format.go
```diff
@@ -0,0 +1,60 @@
...
+			},
+			NewText: newString,
+		},
+	}
+}
+
+func parseSpec(uri, specFile string) (*gauge.Specification, error) {
+	spec, parseResult := new(parser.SpecParser).Parse(getContent(uri), gauge.NewConceptDictionary(), specFile)
+	if !parseResult.Ok {
+		err := parseResult.ParseErrors[0]
```
kashishm: Why is it only returning the first error and not all the errors?

PR #846, opened 2017-11-02
Find references code lens

nehashri:
Adding find references code lens for javascript runner.
- Steps cache caches Steps instead of Step values
- Dynamically registering capabilities for javascript files with lsp client
- Runner is started with gauge daemon in lsp
- Sending javascript file cache requests to runner.
- Added proto message for Step position requests in runner.
- InfoGatherer has new method AllSteps - returns all steps cached, without de-duplicating


api/lang/codeLens.go
```diff
@@ -46,21 +60,44 @@ func getCodeLenses(req *jsonrpc2.Request) (interface{}, error) {
 
 }
 
+func getReferenceCodeLenses(params lsp.CodeLensParams) (interface{}, error) {
+	cacheFileRequest := &gm.Message{MessageType: gm.Message_StepPositionsRequest, StepPositionsRequest: &gm.StepPositionsRequest{FilePath: util.ConvertURItoFilePath(params.TextDocument.URI)}}
```
kashishm: Is this `cacheFileRequest` or `stepPositionsRequest`?

PR #854, opened 2017-11-13
adding custom logger

nehashri:
Custom logger doesn't print to standard output instead sends log message to custom logger
LSP uses this custom logger to send `window/logMessage` to lsp client.

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -40,28 +40,37 @@ const (
 
 var level logging.Level
 var isWindows bool
+var customLogger CustomLogger
+
+type CustomLogger interface {
+	Log(logLevel logging.Level, msg string)
+}
+
+func SetCustomLogger(l CustomLogger) {
+	customLogger = l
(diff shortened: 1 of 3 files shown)


logger/logger.go
```diff
@@ -40,35 +40,52 @@ const (
...
+	Log(logLevel logging.Level, msg string)
+}
+
+func SetCustomLogger(l CustomLogger) {
+	customLogger = l
+}
 
 // Infof logs INFO messages
 func Infof(msg string, args ...interface{}) {
 	GaugeLog.Infof(msg, args...)
```
sriv: perhaps `write` should also include writing to `GaugeLog`.

nehashri: We write to GaugeLog with different log levels. GaugeLog.Infof, GaugeLog.Errorf, etc
Writing to GaugeLog in write method would introduce another switch case.

PR #855, opened 2017-11-13
providing stub for unimplemented step through codeAction

BugDiver:


api/lang/codeAction_test.go
```diff
@@ -0,0 +1,62 @@
...
+	want := []lsp.Command{
+		{
+			Command:   copyStubCommand,
+			Title:     copyStubTitle,
+			Arguments: []interface{}{stub},
+		},
+	}
+
+	got,err := getCodeActions(&jsonrpc2.Request{Params: &p})
+	fmt.Print(got)
```
kashishm: Remove this print statement.

api/lang/diagnostic.go
```diff
@@ -21,7 +21,15 @@ func createDiagnostics(uri string) []lsp.Diagnostic {
 
 func createValidationDiagnostics(errors []validation.StepValidationError, uri string) (diagnostics []lsp.Diagnostic) {
 	for _, err := range errors {
-		diagnostics = append(diagnostics, createDiagnostic(err.Message(), err.Step().LineNo-1, 1, uri))
+		diagnostics = append(diagnostics, lsp.Diagnostic{
```
kashishm: Why can't we use `createDiagnostic` function here?

BugDiver: We are adding one extra field (`Code`) in diagnostic for validation error. `createDiagonastic` doesn't  return a `pointer` (because we don't need one) so we cant change that after creating.

PR #865, opened 2017-11-20
Update to correct Gauge logo

zabil:


zabil:
@haroon-sheikh done. Great catch.

Issue #823, opened 2017-09-27
Show Diagnostics LSP


BugDiver:
Gauge LSP server should support publish diagnostics for spec and concept files.
The server should publish diagnostics when the document changes.
This story is part of [LSP support](https://github.com/getgauge/gauge/issues/717) for gauge.

## Test cases

- [x] Should show critical errors
- [x] Should show spec parse errors
- [x] Should show concept parse errors
- [ ] Should show unimplemented steps
- [ ] duplicate step definition

BugDiver:
Currently diagnostics for error `Should show unimplemented steps`  and ` duplicate step definition` is shown only for Javascript.

Fix should be available with following nightlies =>
 gauge      => 09-1-2017
 gauge-js => 09-1-2017

sguptatw:
Issues
- [ ] Unimplemented step is not getting highlighted
- [ ] Duplicate step implementation is not being shown
- [ ] Number of usages are also not displayed

[test1.zip](https://github.com/getgauge/gauge/files/1469724/test1.zip)

Apoorva-GA:
Please reopen the spec file or edit the spec file for the errors to get highlighted, because diagnostics load when there a file change or when you open the file.

sguptatw:
Closing this issue as the unimplemented step is working as designed.
Unable to replicate the duplicate step issue consistently.

sguptatw:
Getting diagnostics on wrong lines
Getting step implementation not found on the wrong line

<img width="421" alt="parseerrorwrongline" src="https://user-images.githubusercontent.com/6310197/32841709-114d0e1c-ca41-11e7-888c-658a5dc72358.png">

[test1.zip](https://github.com/getgauge/gauge/files/1475077/test1.zip)

sguptatw:
Getting step implementation not found on existing valid steps.
![StepImplementationNotFound.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/571dd284c8c2e113000ddd5a/125516c66e82c728ace21e0d46db809f33ab8eb983ecb246aa49f1802049295d26108524658be76af3006b1300051bff054a5556f9ff8c7faeb76a25dc5f58b8856d5b72c63d15fdbbd11abd13a632c91edbe2e3d7158abd7f37ed998d4d341f001989f85832a6904aca5cfe59ac.png)

sguptatw:
Getting a duplicate step implementation on a step that is not implemented
[...]

PR #867, opened 2017-11-20
publishing diagnostics for all files #823

nehashri:
All files are validated to publish diagnostics.

Diagnostics are published when
- when gauge lsp server is initialized
- a gauge file(.spec and .cpt) is opened
- any file is changed (gauge or implementation files)

A lock on publish diagnostics makes sure only one go routine performs this at a time.

Suggestions are added to duplicate step validation errors. They are filtered before printing suggestions on console.


validation/suggest.go
```diff
@@ -31,13 +31,26 @@ func showSuggestion(validationErrors validationErrors) {
 	if !HideSuggestion {
 		for t, errs := range groupErrors(validationErrors) {
 			fmt.Println(getSuggestionMessage(t))
-			for _, err := range errs {
-				fmt.Println(err.Suggestion())
+			suggestions := filterDuplicateSuggestions(errs)
```
kashishm: When would there be duplicate suggestions?

nehashri: We have duplicate suggestions printed if there is an unimplemented step is used multiple times.

kashishm: I just tried it, it is printing without the duplicate suggestions.

nehashri: In master the suggestions was added to only one error. The duplicate errors did not have the suggestions. But this doesn't work well for lsp diagnostics. We need the suggestions on all errors. Hence, there are changes in validate.go where we copy over the suggestions while duplicating the errors. 
After this change, we had to filter the duplicate suggestion while printing on the console.

PR #870, opened 2017-11-21
Vscode test explorer support

sriv:
for getgauge/gauge-vscode#42

- new request addes `gauge/specs` to retrieve all spec heading in `specs` directory


api/lang/customResponses.go
```diff
@@ -18,6 +39,26 @@ type ScenarioInfo struct {
 	ExecutionIdentifier string `json:"executionIdentifier"`
 }
 
+type specInfo struct {
+	Heading             string `json:"heading"`
+	ExecutionIdentifier string `json:"executionIdentifier"`	
+}
+
+func getSpecs() (interface{}, error) {
```
nehashri: We already have a specs cache in specInfoProvider.
Other lsp features depend on this cache. 
We can use the cache to get all specs.
Any reason we need to parse the files to get all the specs?

api/lang/customResponses.go
```diff
@@ -26,7 +58,17 @@ func getScenarios(req *jsonrpc2.Request) (interface{}, error) {
 		return nil, err
 	}
 	file := util.ConvertURItoFilePath(params.TextDocument.URI)
-	spec, parseResult := new(parser.SpecParser).Parse(getContent(params.TextDocument.URI), gauge.NewConceptDictionary(), file)
+	content:=""
+	if isOpen(params.TextDocument.URI) {
```
nehashri: scenarios can also be fetched from infoProvider.
We don't have to parse a file to do this.

api/lang/server.go
```diff
@@ -145,8 +148,10 @@ func (h *LangHandler) Handle(ctx context.Context, conn jsonrpc2.JSONRPC2, req *j
 		return getCodeActions(req)
 	case "gauge/scenarios":
 		return getScenarios(req)
+	case "gauge/specs":
+		return getSpecs()
 	default:
-		return nil, nil
+		return nil, fmt.Errorf("Unknown request : %s" + req.Method)
```
nehashri: vscode sends a lot of requests like didSave, which we don't make use of. Just because we are not handling these requests, we shouldn't send an error. Also, any errors sent to the client gets printed on the client gauge output panel. It will lead to a lot of noise.

PR #871, opened 2017-11-21
Only allow substitution of variables set outside of gauge

haroon-sheikh:


env/env.go
```diff
@@ -139,14 +135,12 @@ func substituteEnvVars() error {
 				// if not, get from system env
 				envKey, property := match[0], match[1]
 				propertyValue := envVars[property]
```
nehashri: property value is updated regardless. This statement can be removed

nehashri: propertyValue is overridden always. This statement can be removed

haroon-sheikh:
Changes made.

Issue #879, opened 2017-11-28
add support for symbols via LSP


sriv:
### Expected behavior
Add support for 

- [`workspace/symbol`](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#workspace_symbol)
- [`textDocument/documentSymbol`](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_documentSymbol)

This should add support for getgauge/gauge-vscode#63

Document symbols are listed for `spec` and `concepts`.
- `spec` files will have `spec` and `scenario`
- `concept` files will have all concepts

Workspace symbols will list only `spec` and `scenario` from all spec files.

sriv:
should be available in nightly >= 30/11/2017

PR #883, opened 2017-11-29
879 symbols lsp

sriv:
for #879 

- adds documents symbols support
- adds workspace symbols support

### DIFF
--- a/api/lang/symbols.go
+++ b/api/lang/symbols.go
@@ -40,6 +40,9 @@ func documentSymbols(req *jsonrpc2.Request) (interface{}, error) {
 	}
 	file := util.ConvertURItoFilePath(params.TextDocument.URI)
 	content := getContent(params.TextDocument.URI)
+	if util.IsConcept(file) {
+		return getConceptSymbols(content, file), nil
+	}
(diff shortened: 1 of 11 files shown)


api/lang/symbols.go
```diff
@@ -0,0 +1,151 @@
...
+	if err = json.Unmarshal(*req.Params, &params); err != nil {
+		logger.APILog.Debugf("failed to parse request %s", err.Error())
+		return nil, err
+	}
+	file := util.ConvertURItoFilePath(params.TextDocument.URI)
+	content := getContent(params.TextDocument.URI)
+	if util.IsConcept(file) {
+		return getConceptSymbols(content, file), nil
+	}
+	spec, parseResult := new(parser.SpecParser).Parse(content, gauge.NewConceptDictionary(), file)
```
Apoorva-GA: Do we need to parse the file? Can we not get it from the cache?
NOTE : InCase of workspace symbols we are getting it from cache.

sriv: `documentSymbols` are generated from buffer content, so need to be parsed again. (Since the document can be opened and edited, without saving).

If we enable autosave, we do not need this.

PR #891, opened 2017-12-04
Adding hook messages after hook execution. getgauge/html-report#111

nehashri:


execution/result/suiteResult.go
```diff
@@ -40,6 +40,8 @@ type SuiteResult struct {
 	ProjectName       string
 	Timestamp         string
 	SpecsSkippedCount int
+	PreHookMessage    []string
```
sriv: Should be `PreHookMessages` and `PostHookMessages` (plural), since this is an array. This is particularly important in the contract since proto would generate methods like `GetPreHookMessage` which indicates a singular value.

Issue #862, opened 2017-11-17
Executing the previous run from command line


BugDiver:
User should be able to run the previous run from the command line.

**Proposed behavior**
* Add a flag `--repeat` to `run` subcommand.
* Execute the last run with same flags and env by running `gauge run --repeat` command.

haroon-sheikh:
@BugDiver @riju91  What is the use case of this? 

It sounds like a good feature only if we could pass the number of times to repeat. This can be really useful when trying to replicate a intermitted issue.

E.g. `gauge run --repeat=3`. 

I guess the only problem will be reporting when running multiple times.

sriv:
@haroon-sheikh - this issue is a pre-requisite for https://github.com/getgauge/gauge-vscode/issues/41

TLDR: we are implementing a feature in vscode plugin that will allow users to author implementation, and repeat scenario run without having to toggle between files.

BugDiver:
should be available in nightly >= 20-11-2017

sguptatw:
# Test cases
- [ ] Should be able to repeat run for passed, failed, skipped tests
- [ ] Should give an error message if there is no last run
- [ ] Should be able to repeat run with tags, parallel run, data driven execution, selective data driven execution, folder execution

sguptatw:
If gauge does not execute specs
example: - 
- [ ] No specifications found in specs. - for a tagged expression
- [ ] Failed to load env. something1 environment does not exist - for a non existing env input

What should be the behavior of re-run in such scenarios

zabil:
> What should be the behavior of re-run in such scenarios

It should show and info on the problems mentioned.

sguptatw:
**Expected behavior**
Should show info when previous run has failed.

**Actual behavior**
Runs the previously successful gauge execution.

BugDiver:
Should be fixed in nightly >= 27-11-2017

sguptatw:
Please find the failing test cases inline in the previous comment

sguptatw:
If the previous command is `gauge run --failed` what should `gauge run --repeat` do?

sriv:
> If the previous command is gauge run --failed what should gauge run --repeat do when run multiple times? Should it run only the failed tests repeatedly?

IMO, `gauge run --repeat` should _just_ repeat the last run. It should be synonymous with running the previous command again.

In this case this:

```
> gauge run --failed
> gauge run --repeat
> gauge run --repeat
```
should be equivalent to:

```
gauge run --failed
[...]

PR #894, opened 2017-12-07
Adding Repeat Last Command Feature #862

riju91:
* Reverted the previous implementation which was saving Items
* We are saving the previously executed command given the user for repeat

### DIFF
--- a/api/lang/customResponses_test.go
+++ b/api/lang/customResponses_test.go
@@ -72,7 +72,7 @@ Scenario Heading2
 	f.remove(uri)
 }
 
-func Test_getScenariosShouldGiveTheScenariosIfCursorPositionIsNotInSpan(t *testing.T) {
+func TestGetScenariosShouldGiveTheScenariosIfCursorPositionIsNotInSpan(t *testing.T) {
 	specText := `Specification Heading
 =====================
 
--- a/api/lang/customResponses_test.go
+++ b/api/lang/customResponses_test.go
@@ -72,8 +72,7 @@ Scenario Heading2
 	f.remove(uri)
 }
 
-func TestGetScenariosShouldGiveTheScenariosIfCursorPositionIsNotInSpan(t *testing.T) {
-	provider = &dummyInfoProvider{}
+func Test_getScenariosShouldGiveTheScenariosIfCursorPositionIsNotInSpan(t *testing.T) {
 	specText := `Specification Heading
 =====================
 
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -20,6 +20,11 @@ package cmd
 import (
 	"fmt"
 
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
 	"github.com/getgauge/common"
 	"github.com/getgauge/gauge/config"
 	"github.com/getgauge/gauge/env"
@@ -34,6 +39,27 @@ import (
 	"github.com/getgauge/gauge/validation"
 	"github.com/spf13/cobra"
 )
+
+const (
+	dotGauge        = ".gauge"
+	prevCmdFileName = "prevCmd.json"
+)
+
+type prevCommand struct {
+	Command []string
+}
+
+func newPrevCommand() *prevCommand {
+	return &prevCommand{Command: make([]string, 0)}
+}
+
+func getJSON(failedMeta *prevCommand) string {
+	j, err := json.MarshalIndent(failedMeta, "", "\t")
+	if err != nil {
+		logger.Warningf("Failed to save run info. Reason: %s", err.Error())
+	}
+	return string(j)
+}
 
 var (
 	GaugeCmd = &cobra.Command{
@@ -51,6 +77,7 @@ var (
 		},
 		DisableAutoGenTag: true,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			handleRepeatCommand(cmd)
 			skel.CreateSkelFilesIfRequired()
 			track.Init()
 			config.SetProjectRoot(args)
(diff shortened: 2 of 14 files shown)


cmd/cmd.go
```diff
@@ -35,6 +40,27 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	dotGauge        = ".gauge"
```
nehashri: There is already a constant for .gauge folder defined in common. We should be using that variable everywhere.

cmd/cmd.go
```diff
@@ -35,6 +40,27 @@ import (
...
+	prevCmdFileName = "prevCmd.json"
```
nehashri: `prevCmd.json` filename is a little ambiguous. It can just mean previous gauge command. Can rename it to something else? like `lastRunCmd`?

cmd/cmd.go
```diff
@@ -35,6 +40,27 @@ import (
...
+
+type prevCommand struct {
+	Command []string
+}
+
+func newPrevCommand() *prevCommand {
+	return &prevCommand{Command: make([]string, 0)}
+}
+
+func getJSON(failedMeta *prevCommand) string {
```
nehashri: can be a function on `prevCommand` struct itself.

cmd/cmd.go
```diff
@@ -35,6 +40,27 @@ import (
...
+	j, err := json.MarshalIndent(failedMeta, "", "\t")
+	if err != nil {
+		logger.Warningf("Failed to save run info. Reason: %s", err.Error())
```
nehashri: Doesn't look like the method is saving the run info here. Instead we can return the error here. and use the appropriate log message where needed.

execution/rerun/rerun.go
```diff
@@ -37,32 +37,27 @@ import (
 )
 
 const (
-	dotGauge     = ".gauge"
-	infoFileName = "runInfo.json"
+	dotGauge   = ".gauge"
```
nehashri: There is already a constant for .gauge folder defined in common. We should be using that variable everywhere.

cmd/run.go
```diff
@@ -78,19 +78,25 @@ func init() {
 	runCmd.Flags().StringVarP(&strategy, "strategy", "", "lazy", "Set the parallelization strategy for execution. Possible options are: `eager`, `lazy`")
 	runCmd.Flags().BoolVarP(&sort, "sort", "s", false, "Run specs in Alphabetical Order")
 	runCmd.Flags().BoolVarP(&failed, "failed", "f", false, "Run only the scenarios failed in previous run")
-	runCmd.Flags().BoolVarP(&repeat, "repeat", "", false, "Repeat the previous run")
+	runCmd.Flags().BoolVarP(&repeat, "repeat", "", false, "Repeat last command")
 	runCmd.Flags().BoolVarP(&hideSuggestion, "hide-suggestion", "", false, "Prints a step implementation stub for every unimplemented step")
 }
 
+//This flag stores whether the command is gauge run --failed and if it is triggering another command.
+//The purpose is to only store commands given by user in the runInfo file.
```
nehashri: runInfo file?

cmd/cmd.go
```diff
@@ -67,6 +94,56 @@ var (
...
+
+func executeCmd(cmd *cobra.Command, lastState []string) {
+	cmd.Parent().SetArgs(lastState[1:])
+	os.Args = lastState
+	resetFlags()
+	cmd.Execute()
+}
+
+func readPrevCmd() *prevCommand {
+	contents, err := common.ReadFileContents(filepath.Join(config.ProjectRoot, rerun.DotGauge, lastRunCmdFileName))
```
nehashri: There is already a constant for .gauge folder defined in common. We should be using that variable everywhere.

cmd/cmd.go
```diff
@@ -67,6 +94,56 @@ var (
...
+
+func writePrevCmd() {
+	prevCmd := newPrevCommand()
+	prevCmd.Command = os.Args
+	contents, err := prevCmd.getJSON()
+	if err != nil {
+		logger.Fatalf("Unable to parse last run command. Error : %v", err.Error())
+	}
+	prevCmdFile := filepath.Join(config.ProjectRoot, rerun.DotGauge, lastRunCmdFileName)
+	dotGaugeDir := filepath.Join(config.ProjectRoot, rerun.DotGauge)
```
nehashri: There is already a constant for .gauge folder defined in common. We should be using that variable everywhere.

execution/rerun/rerun.go
```diff
@@ -37,32 +37,27 @@ import (
 )
 
 const (
-	dotGauge     = ".gauge"
-	infoFileName = "runInfo.json"
+	DotGauge   = ".gauge"
```
nehashri: There is already a constant for .gauge folder defined in common. We should be using that variable everywhere.
.gauge folder can be used in other packages too. So why export it here?

execution/rerun/rerun.go
```diff
@@ -145,10 +141,9 @@ func addFailedMetadata(res result.Result, args []string, add func(res result.Res
 	}
 }
 
-func WriteLastRunInfo() {
-	contents := getJSON(runInfo)
-	failuresFile := filepath.Join(config.ProjectRoot, dotGauge, infoFileName)
-	dotGaugeDir := filepath.Join(config.ProjectRoot, dotGauge)
+func writeFailedMeta(contents string) {
+	failuresFile := filepath.Join(config.ProjectRoot, DotGauge, failedFile)
```
nehashri: There is already a constant for .gauge folder defined in common. We should be using that variable everywhere.

api/lang/customResponses_test.go
```diff
@@ -72,8 +72,7 @@ Scenario Heading2
 	f.remove(uri)
 }
 
-func TestGetScenariosShouldGiveTheScenariosIfCursorPositionIsNotInSpan(t *testing.T) {
-	provider = &dummyInfoProvider{}
+func Test_getScenariosShouldGiveTheScenariosIfCursorPositionIsNotInSpan(t *testing.T) {
```
BugDiver: use the existing naming format for tests

cmd/cmd.go
```diff
@@ -35,6 +40,26 @@ import (
...
+type prevCommand struct {
+	Command []string
+}
+
+func newPrevCommand() *prevCommand {
+	return &prevCommand{Command: make([]string, 0)}
+}
+
+func (failedMeta *prevCommand) getJSON() (string, error) {
+	j, err := json.MarshalIndent(failedMeta, "", "\t")
```
BugDiver: change the name `failedMeta` to something appropriate

cmd/cmd.go
```diff
@@ -51,6 +77,7 @@ var (
 		},
 		DisableAutoGenTag: true,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			handleRepeatCommand(cmd)
```
BugDiver: It's gonna capture all commands. Example:- if I run `gauge install java`, it will store that in the  .gauge dir.
`repeat` is very specific to `run` and should not be capturing all the commands.

cmd/cmd.go
```diff
@@ -67,6 +93,58 @@ var (
...
+func handleRepeatCommand(cmd *cobra.Command, cmdArgs []string) {
+	if repeat {
+		prevCmd := readPrevCmd()
+		executeCmd(cmd, prevCmd.Command)
+	} else {
+		if prevFailed {
+			prevFailed = false
+			return
+		}
+		if cmd.Name() == "run" {
```
nehashri: Can we move this to `PreRun` on the run command? 
This check for the command name explicitly is not needed then.

cmd/docs.go
```diff
@@ -32,6 +33,9 @@ var docsCmd = &cobra.Command{
 	Long:    `Generate documentation using specified plugin.`,
 	Example: "  gauge docs spectacle specs/",
 	Run: func(cmd *cobra.Command, args []string) {
+		if e := env.LoadEnv(environment); e != nil {
```
nehashri: This is not a project level command. We do not need to load the env.

cmd/cmd.go
```diff
@@ -67,6 +65,13 @@ var (
 	gaugeVersion    bool
 )
 
+func executeCmd(cmd *cobra.Command, lastState []string) {
```
nehashri: can be moved to run.go file

cmd/run.go
```diff
@@ -108,6 +143,55 @@ func execute(args []string) {
 		logger.Fatalf(e.Error())
 	}
 	exitCode := execution.ExecuteSpecs(specs)
-	rerun.WriteLastRunInfo()
 	os.Exit(exitCode)
 }
+
+func handleRepeatCommand(cmd *cobra.Command, cmdArgs []string) {
```
nehashri: Test for this function?

PR #895, opened 2017-12-08
using channels to return right response to caller getgauge/gauge-vscode#54

BugDiver:


conn/network.go
```diff
@@ -19,23 +19,50 @@ package conn
...
 	"time"
 
 	"github.com/getgauge/common"
 	"github.com/getgauge/gauge/gauge_messages"
 	"github.com/golang/protobuf/proto"
 )
 
-func WriteDataAndGetResponse(conn net.Conn, messageBytes []byte) ([]byte, error) {
+type response struct {
+	result chan *gauge_messages.Message
```
sriv: can `result` and `err` be directional channel?

conn/network_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if err != nil {
+		t.Errorf("expected err to be nil. got %v", err)
+	}
+	if !reflect.DeepEqual(res, responseMessage) {
+		t.Errorf("expected : %v\ngot : %v", responseMessage, res)
+	}
+}
+
+func TestGetResponseForGaugeMessageShoudGiveTheRightResponse(t *testing.T) {
```
sriv: There is only one request/response. So this test possibly cannot fail?

Issue #890, opened 2017-12-04
Renaming a spec file increases the find usages count


sguptatw:
**Expected behavior**
Renaming of spec files should not impact the usage count

**Actual behavior**
The count increases.

**Steps to replicate**
* Create a gauge project
* Rename a spec file
* Now goto the definition of a step

> Observe the usage count shown is wrong

**Version**
```
Gauge version: 0.9.6.nightly-2017-12-01
Commit Hash: fdcd4e2

Plugins
-------
js (2.0.3.nightly-2017-11-30)
```
IDE - gauge-0.0.3.nightly-2017-12-01.vsix

BugDiver:
All the caches in specInfoGatherer are not updated when a specification file is removed or renamed.
The same issue can occur for tags and params completion.

BugDiver:
should be fixed in nightly >= 13-12-2017

PR #896, opened 2017-12-08
Updating steps cache on file delete #890

riju91:


riju91:
should be available in nightly >= 12-12-2017

Issue #902, opened 2017-12-19
Peek definition of a step does not highlight the step definition


Apoorva-GA:
_From @zabil on December 13, 2017 11:12_

```
VS-Code version: 1.18.1
Gauge Extension Version: 0.0.3

Gauge version: 0.9.6
Commit Hash: 0b85c7e

Plugins
-------
csharp (0.10.1)
go (0.1.1)
html-report (4.0.2)
java (0.6.5.nightly-2017-09-05)
js (2.0.3)
python (0.2.2)
ruby (0.4.2)
screenshot (0.0.1)

```

### Steps to replicate
```
$ gauge init js
$ code .
```

* Open `example.spec` in VS code
* Use peek definition (ctrl+shift+f10) on any step
* Step implementation is not highlighted

![peekdefintion](https://user-images.githubusercontent.com/54427/33936054-24268b64-e024-11e7-934b-582f1fad757f.png)

### Expected behavior
Highlight the function or function name for e.g.

![highlightedreference](https://user-images.githubusercontent.com/54427/33936143-81c84a64-e024-11e7-8f84-0656df3bc020.png)

_Copied from original issue: getgauge/gauge-vscode#72_

Apoorva-GA:
On peek definition the entire step definition gets highlighted.

Should be available in gauge and gauge-js nightly >= 21/12/17

sguptatw:
Getting `Duplicate step implementation` on every step after clicking on `Peek Definition`.

Apoorva-GA:
Fix should be available in gauge-js nightly >= 8/1/18

PR #903, opened 2017-12-19
Highlighting the step when showing the peek definition #902

Apoorva-GA:


api/lang/definition.go
```diff
@@ -86,22 +87,48 @@ func searchStep(step *gauge.Step) (interface{}, error) {
...
+
+func getLspLocationForConcept(fileName string, lineNumber int) (interface{}, error) {
+	uri := util.ConvertPathToURI(fileName)
+	var endPos int
+	diskFileCache := &files{cache: make(map[string][]string)}
+	lineNo := lineNumber - 1
+	if isOpen(uri) {
+		endPos = len(getLine(uri, lineNo))
+	} else {
+		if !diskFileCache.exists(uri) {
```
BugDiver: Why do we need this check? This condition is always gonna be true

BugDiver: Why do we need this check? This condition is always gonna be true

PR #910, opened 2018-01-05
 Plugin requests during execution do not timeout

Apoorva-GA:


conn/network.go
```diff
@@ -34,6 +34,21 @@ import (
...
+func (r *response) stopTimer() {
+	if r.timer != nil {
+		r.timer.Stop()
+	}
+}
+
+func (r *response) addTimer(timeout time.Duration, messageType string) {
+	if timeout > 0 {
+		r.timer = time.AfterFunc(timeout, func() {
+			r.err <- fmt.Errorf("%s request timed out.", messageType)
```
sriv: Should perhaps also have the message ID. The older message was showing which message timed out, which is useful when looking at API logs.

PR #927, opened 2018-01-25
logEnv fix

lbarzolao:


nehashri:
@lbarzolao, thanks for the PR. But is there any specific error you are facing that the PR is trying to fix? 
You seem to have exported a variable which is not used outside of the package. This is not necessary if the variable is used only within the package.

lbarzolao:
The issue i facing is i not able to set up a custom logs_directory. I looked in env.go and there is set a **LogsDirectory** env. variable, but logger.go is trying to get a **logsDirectory** env. variable.

nehashri:
@lbarzolao Both the variables in logger.go and env.go have the same value `logs_directory` 
`logsDirectory    = "logs_directory"` in logger.go
`LogsDirectory       = "logs_directory"` in env.go

The variables seem to be duplicated across files, none the less they are assigned the same value `logs_directory`.
Did you try and install gauge from source after your fix? 
There seems to be a issue with the custom logs directory, but the above fix doesn't seem to fix it...

nehashri:
Closing this issue.

PR #930, opened 2018-01-31
Adding codelens for debug

Apoorva-GA:


api/lang/codeLens_test.go
```diff
@@ -59,6 +59,18 @@ Scenario Heading
...
 
+	specDebugCodeLens := lsp.CodeLens{
+		Command: lsp.Command{
+			Command:   "gauge.debug",
+			Title:     "Debug Spec",
+			Arguments: getExecutionArgs("foo.spec"),
+		},
+		Range: lsp.Range{
+			Start: lsp.Position{0, 0},
+			End:   lsp.Position{0, 10},
```
riju91: As per protocol, this should be `The range in which this code lens is valid` as documented in https://microsoft.github.io/language-server-protocol/specification#textDocument_codeLens

Apoorva-GA: Yeah, but this will be problem with the references code lens. Where we don't have the info about the length of the line.

PR #935, opened 2018-02-06
Sending execution status info to vscode

Apoorva-GA:


execution/execute.go
```diff
@@ -159,6 +167,67 @@ func newExecution(executionInfo *executionInfo) suiteExecutor {
...
+	SpecsPassed   int
+	SpecsFailed   int
+	SpecsSkipped  int
+	SceExecuted   int
+	ScePassed     int
+	SceFailed     int
+	SceSkipped    int
+}
+
+func newExecutionStatus() *executionStatus {
```
BugDiver: No need for the explicit assignment. Integer's default value is 0.

execution/execute.go
```diff
@@ -159,6 +167,67 @@ func newExecution(executionInfo *executionInfo) suiteExecutor {
...
+	return &executionStatus{SpecsExecuted: 0, SpecsPassed: 0, SpecsFailed: 0, SpecsSkipped: 0, SceExecuted: 0, ScePassed: 0, SceFailed: 0, SceSkipped: 0}
+}
```
BugDiver: No need for the explicit assignment. Integer's default value is 0.

execution/execute.go
```diff
@@ -159,6 +167,67 @@ func newExecution(executionInfo *executionInfo) suiteExecutor {
...
+
+func (status *executionStatus) getJSON() (string, error) {
+	j, err := json.MarshalIndent(status, "", "\t")
+	if err != nil {
+		return "", err
+	}
+	return string(j), nil
+}
+
+func writeExecutionStatus(executedSpecs int, passedSpecs int, failedSpecs int, skippedSpecs int, executedScenarios int, passedScenarios int, failedScenarios int, skippedScenarios int) {
```
BugDiver: Don't need to specify the type for all params if the type is same. It can be `function(a,b,c int)`

execution/execute.go
```diff
@@ -159,6 +167,67 @@ func newExecution(executionInfo *executionInfo) suiteExecutor {
...
+	SpecsExecuted int
+	SpecsPassed   int
+	SpecsFailed   int
+	SpecsSkipped  int
+	SceExecuted   int
+	ScePassed     int
+	SceFailed     int
+	SceSkipped    int
+}
+
```
BugDiver: This struct will be marshaled to JSON. Use `json` tags

PR #936, opened 2018-02-07
119 generate stub

riju91:
2 API added
1. Get a list of implementation files (from language runner)
2. Given a file and a code snippet, put the code snippet in appropriate location and return new content (with the help of language runner)


api/lang/customResponses.go
```diff
@@ -49,6 +55,43 @@ func specs() (interface{}, error) {
...
+func putStubImpl(req *jsonrpc2.Request) (interface{}, error) {
+	var stubImplParams StubImpl
+	if err := json.Unmarshal(*req.Params, &stubImplParams); err != nil {
+		logger.APILog.Debugf("failed to parse request %s", err.Error())
+		return nil, err
+	}
+	if lRunner.runner == nil {
+		return nil, nil
+	}
+	var stepVals []*gm.StepValidateRequest
```
nehashri: Not a good idea to use step validate request here.

api/lang/customResponses.go
```diff
@@ -40,6 +41,11 @@ type specInfo struct {
 	ExecutionIdentifier string `json:"executionIdentifier"`
 }
 
+type StubImpl struct {
```
nehashri: Does this struct need to be exported?

api/lang/customResponses.go
```diff
@@ -49,6 +55,43 @@ func specs() (interface{}, error) {
...
+		}
+		protoStepValue := gauge.ConvertToProtoStepValue(stepValue)
+
+		stepVals = append(stepVals, protoStepValue)
+	}
+	putStubImplementationResponse, err := putStubImplementation(stubImplParams.ImplementationFilePath, stepVals)
+	if err != nil {
+		return nil, err
+	}
+	return putStubImplementationResponse, nil
```
nehashri: A list of [workspaceEdits](https://microsoft.github.io/language-server-protocol/specification#workspaceedit) should be sent back to VSCode. In LSP, All changes that need to be done on the files should be done with a workspaceEdit.

api/lang/runner.go
```diff
@@ -98,6 +101,48 @@ func getStepPositionResponse(uri string) (*gm.StepPositionsResponse, error) {
...
+
+func putStubImplementation(result *lsp.WorkspaceEdit, filePath string, stepTexts []*gm.ProtoStepValue) error {
+	stubImplementationCodeRequest := &gm.Message{MessageType: gm.Message_StubImplementationCodeRequest, StubImplementationCodeRequest: &gm.StubImplementationCodeRequest{ImplementationFilePath: filePath, Steps: stepTexts}}
+	response, err := GetResponseFromRunner(stubImplementationCodeRequest)
+	if err != nil {
+		logger.APILog.Infof("Error while connecting to runner : %s", err.Error())
+		return err
+	}
+	fileEditResponse := response.GetFileChanges()
+	uri := util.ConvertPathToURI(fileEditResponse.FileName)
```
nehashri: It is a good idea to keep the extra manipulations done to runner responses out of this file. Please move the logic to create workspaceEdits out of this file. If possible, reuse already existing methods to do this.

PR #952, opened 2018-02-26
Get language id for debugger, getgauge/gauge-python#33

riju91:


api/lang/customResponses.go
```diff
@@ -84,6 +84,10 @@ func putStubImpl(req *jsonrpc2.Request) (interface{}, error) {
 	return getWorkspaceEditForStubImpl(fileChanges, stubImplParams.ImplementationFilePath), nil
 }
 
+func getRunnerLanguage() (interface{}, error) {
+	return lRunner.lspID, nil
+}
```
BugDiver: this method has only one line. can it be inlined?

api/lang/server.go
```diff
@@ -180,6 +180,8 @@ func (h *LangHandler) Handle(ctx context.Context, conn jsonrpc2.JSONRPC2, req *j
 		return specs()
 	case "gauge/executionStatus":
 		return execution.ReadExecutionStatus()
+	case "gauge/getRunnerLanguage":
+		return getRunnerLanguage()
```
BugDiver: follow the same convention as other requests. Can it be `gauge/runnerLanguage`  or some other name?

PR #957, opened 2018-03-02
Better logger

sriv:
refactor logger package,
trace lsp requestid, method so that it can be correlated better.

### DIFF
--- a/api/lang/completionStep.go
+++ b/api/lang/completionStep.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 
 	"github.com/getgauge/gauge/gauge"
-	"github.com/getgauge/gauge/logger"
 	"github.com/getgauge/gauge/parser"
 	"github.com/sourcegraph/go-langserver/pkg/lsp"
 )
@@ -32,9 +31,7 @@ func stepCompletion(line, pLine string, params lsp.TextDocumentPositionParams) (
 	list := completionList{IsIncomplete: false, Items: make([]completionItem, 0)}
 	editRange := getStepEditRange(line, params.Position)
 	prefix := getPrefix(pLine)
-	var givenArgs []gauge.StepArg
-	var err error
-	givenArgs, err = getStepArgs(strings.TrimSpace(pLine))
+	givenArgs, err := getStepArgs(strings.TrimSpace(pLine))
 	if err != nil {
 		return nil, err
 	}
@@ -43,13 +40,14 @@ func stepCompletion(line, pLine string, params lsp.TextDocumentPositionParams) (
 		cText := prefix + addPlaceHolders(c.StepValue.StepValue, c.StepValue.Parameters)
 		list.Items = append(list.Items, newStepCompletionItem(c.StepValue.ParameterizedStepValue, cText, concept, fText, editRange))
 	}
-	allSteps := append(allUsedStepValues(), allImplementedStepValues()...)
+	s, err := allImplementedStepValues()
+	allSteps := append(allUsedStepValues(), s...)
 	for _, sv := range removeDuplicates(allSteps) {
 		fText := prefix + getStepFilterText(sv.StepValue, sv.Args, givenArgs)
 		cText := prefix + addPlaceHolders(sv.StepValue, sv.Args)
 		list.Items = append(list.Items, newStepCompletionItem(sv.ParameterizedStepValue, cText, step, fText, editRange))
 	}
-	return list, nil
+	return list, err
 }
 
 func removeDuplicates(steps []gauge.StepValue) []gauge.StepValue {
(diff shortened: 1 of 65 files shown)


logger/logger.go
```diff
@@ -34,17 +34,29 @@ import (
...
 
 var level logging.Level
-var isWindows bool
 var customLogger CustomLogger
+var activeLogger *logging.Logger
+var fileLogFormat = logging.MustStringFormatter("%{time:15:04:05.000} %{message}")
+var isLSP bool
+var initialized bool
 
 type CustomLogger interface {
```
nehashri: This interface can be removed now that lang package invokes logger through its own logger. The lang package can do any custom logging it needs on its own.

api/lang/completionStep.go
```diff
@@ -75,7 +75,7 @@ func allImplementedStepValues() []gauge.StepValue {
 	var stepValues []gauge.StepValue
 	res, err := getAllStepsResponse()
 	if err != nil {
-		logger.APILog.Debugf("failed to get steps from runner. %v", err.Error())
+		logger.Debugf(false, "failed to get steps from runner. %v", err.Error())
```
nehashri: should use lang package's logger.

api/lang/diagnostics.go
```diff
@@ -51,7 +51,7 @@ func publishDiagnostics(ctx context.Context, conn jsonrpc2.JSONRPC2) {
 
 		diagnosticsMap, err := getDiagnostics()
 		if err != nil {
-			logger.Errorf("Unable to publish diagnostics, error : %s", err.Error())
+			logger.Errorf(true,"Unable to publish diagnostics, error : %s", err.Error())
```
nehashri: should use lang package's logger

api/lang/diagnostics.go
```diff
@@ -132,7 +132,7 @@ func validateConcepts(diagnostics map[lsp.DocumentURI][]lsp.Diagnostic) (*gauge.
 		}
 		content, err := getContentFromFileOrDisk(conceptFile)
 		if err != nil {
-			logger.Errorf("Unable to read file %s", err)
+			logger.Errorf(true,"Unable to read file %s", err)
```
nehashri: should use lang package's logger

api/lang/format.go
```diff
@@ -36,7 +36,7 @@ func format(request *jsonrpc2.Request) (interface{}, error) {
 	if err := json.Unmarshal(*request.Params, &params); err != nil {
 		return nil, err
 	}
-	logger.APILog.Debugf("LangServer: request received : Type: Format Document URI: %s", params.TextDocument.URI)
+	logger.Debugf(false, "LangServer: request received : Type: Format Document URI: %s", params.TextDocument.URI)
```
nehashri: should use lang package's logger

api/lang/rename.go
```diff
@@ -62,7 +62,7 @@ func rename(ctx context.Context, conn jsonrpc2.JSONRPC2, req *jsonrpc2.Request)
 
 	refactortingResult := refactor.GetRefactoringChanges(step.GetLineText(), newName, lRunner.runner, []string{common.SpecsDirectoryName})
 	for _, warning := range refactortingResult.Warnings {
-		logger.Warningf(warning)
+		logger.Warningf(true, warning)
```
nehashri: should use lang package's logger

api/lang/rename.go
```diff
@@ -40,7 +40,7 @@ func rename(ctx context.Context, conn jsonrpc2.JSONRPC2, req *jsonrpc2.Request)
 	var params lsp.RenameParams
 	var err error
 	if err = json.Unmarshal(*req.Params, &params); err != nil {
-		logger.APILog.Debugf("failed to parse rename request %s", err.Error())
+		logger.Debugf(false, "failed to parse rename request %s", err.Error())
```
nehashri: should use lang package's logger

api/lang/runner.go
```diff
@@ -52,32 +52,29 @@ func startRunner() error {
 }
 
 func connectToRunner(killChan chan bool) (runner.Runner, error) {
-	logger.GaugeLog.Infof("Starting language runner")
+	logger.Infof(false, "Starting language runner")
```
nehashri: should use lang package's logger

api/lang/completionStep.go
```diff
@@ -43,13 +40,14 @@ func stepCompletion(line, pLine string, params lsp.TextDocumentPositionParams) (
 		cText := prefix + addPlaceHolders(c.StepValue.StepValue, c.StepValue.Parameters)
 		list.Items = append(list.Items, newStepCompletionItem(c.StepValue.ParameterizedStepValue, cText, concept, fText, editRange))
 	}
-	allSteps := append(allUsedStepValues(), allImplementedStepValues()...)
+	s, err := allImplementedStepValues()
```
nehashri: An error is thrown if the runner is not up yet. This change would mean that an error is sent back to the client. So if no runner is started, autocomplete of steps may always throw an error and not  show auto complete steps to the user. I am not sure if this is what is expected here

sriv: @nehashri - it should be handled via https://github.com/getgauge/gauge/pull/957/commits/b40f5d95610d04fbc41b5d8eefe8b2b072ad06fe#diff-0cdee1df6466adfecc39d07cef7c843fR71

Issue #973, opened 2018-03-12
Codesign gauge MacOS binaries


BugDiver:
_From @zabil on March 7, 2018 10:21_

### Current behavior
* Whenever the gauge daemon runs (for e.g from VS Code) it pops up the "Allow connections" dialog in macOS
* Not allowing the daemon to open connections and breaks functionality.

### Expected behavior
* Whitelist gauge binary on MacOS firewall.
* codesign the MacOS binary to prevent popups

More information here https://support.apple.com/en-in/HT201642

_Copied from original issue: getgauge/gauge-vscode#184_

BugDiver:
The fix should be available in nightly 2018-03-16

BugDiver:
The zip packages contain signed binaries. The pop up issue should be fixed through https://github.com/getgauge/gauge/commit/49bc97d65ddd6de14887654814ace825ede720f1

PR #974, opened 2018-03-12
signing gauge binaries for mac os #973

BugDiver:
removing pkg installers.


build/make.go
```diff
@@ -171,7 +166,7 @@ var (
 		map[string]string{GOARCH: X86, GOOS: windows, CC: "i586-mingw32-gcc", CGO_ENABLED: "1"},
 		map[string]string{GOARCH: X86_64, GOOS: windows, CC: "x86_64-w64-mingw32-gcc", CGO_ENABLED: "1"},
 	}
-	osDistroMap = map[string]distroFunc{windows: createWindowsDistro, linux: createLinuxPackage, darwin: createDarwinPackage}
+	osDistroMap = map[string]distroFunc{windows: createWindowsDistro, linux: createDarwinPackage, darwin: createDarwinPackage}
```
sriv: we should have a separate `createLinuxPackage` func. This is confusing.

build/make.go
```diff
@@ -105,6 +90,16 @@ func compileGauge() {
 	runProcess("go", args...)
 }
 
+func signDarwinExecutable(execPath string) {
+	id := os.Getenv("OS_SIGNING_IDENTITY")
+	if id == "" {
+		log.Printf("No singning identity found . Executable won't be signed.")
+	}
+	if getGOOS() == "darwin" {
```
sriv: This can be the first check. No need to check for `OS_SIGNING_IDENTITY` if it is not `darwin` right?

PR #979, opened 2018-03-16
Cache file request to send file status events to runner.

nehashri:


api/lang/document.go
```diff
@@ -83,13 +94,57 @@ func documentClosed(req *jsonrpc2.Request, ctx context.Context, conn jsonrpc2.JS
...
 	}
 	go publishDiagnostics(ctx, conn)
 	return err
 }
+
+func documentCreate(req *jsonrpc2.Request, ctx context.Context, conn jsonrpc2.JSONRPC2) error {
+	var params lsp.TextDocumentIdentifier
+	var err error
+	if err = json.Unmarshal(*req.Params, &params); err != nil {
+		return fmt.Errorf("failed to parse request %v", err)
```
sriv: let's not do `%v`, the error messages are unreadable since it brings field names etc. `%s` should be good enough.

api/lang/document.go
```diff
@@ -83,13 +94,57 @@ func documentClosed(req *jsonrpc2.Request, ctx context.Context, conn jsonrpc2.JS
...
+	}
+	go publishDiagnostics(ctx, conn)
+	return err
+}
+
+func documentDelete(req *jsonrpc2.Request, ctx context.Context, conn jsonrpc2.JSONRPC2) error {
+	var params lsp.TextDocumentIdentifier
+	var err error
+	if err := json.Unmarshal(*req.Params, &params); err != nil {
+		return fmt.Errorf("failed to parse request. %v", err)
```
sriv: same as before, `%s` instead of `%v`

Issue #984, opened 2018-03-22
[Feature] Allow to set specification folder as env variable, such as reports directory


jackperalta:
### Expected behavior
It should be possible to define the location of the specification as env variable in the properties file or via command line interface. This allows for customized spec locations to make it easier to execute them, by not forcing anyone to give the custom location with each run.
Especially when one handles multiple projects with different custom locations.
It must be guaranteed, that current behavior is not changed, so that if no env variable is given, the current default `specs` is used.
Also when given a location via command line, e.g. `gauge run somewhereElse`, that this is selected over a variable defined in the properties files

### Actual behavior
Currently one cannot define such env variable

nehashri:
should be available in nightly >=  27-03-18

nehashri:
More details on the feature: 
There can be multiple spec directories in a project. Providing a way to specify this for all runs is useful as the user can then run only `gauge run` command. 
An env variable has been introduced `gauge_specs_dir` which allows a user to specify a comma separated list of spec directories. If the env variable is not specified the directory `specs` will be considered by default.

sguptatw:
**Question**
What should be the behavior of `gauge run` when gauge_specs_dir has multiple spec directories given?

sriv:
>  What should be the behavior of `gauge run` when `gauge_specs_dir` has multiple spec directories given?

I guess it should run all the specs from all the directories specified. That is the behaviour if you specify multiple specs directory from cli. ex `gauge run specs specs1` runs specs from both `specs` and `specs1`

PR #985, opened 2018-03-22
Specs dir can now be defined as env variable

jackperalta:
(my very first) pull request for https://github.com/getgauge/gauge/issues/984

Comments are welcomed


vendor/github.com/getgauge/common/common.go
```diff
@@ -50,7 +50,7 @@ const (
 	ProductName             = "gauge"
 	DotGauge                = ".gauge"
 	config                  = "config"
-	SpecsDirectoryName      = "specs"
+	//SpecsDirectoryName      = "specs"
```
nehashri: It is useful to have this constant. 
Also, In golang vendor folder is used to store external dependencies. So a changes here should not be done.

env/env.go
```diff
@@ -42,6 +43,17 @@ var envVars map[string]string
 
 var currentEnv = "default"
 
+// GetSpecDir returns the specification directory.
+// It checks whether the environment variable for gauge_specs_dir is set.
+// It returns 'specs' otherwise
+func GetSpecDir() string {
```
nehashri: Would be better if this was moved to util.go.

env/env.go
```diff
@@ -42,6 +43,17 @@ var envVars map[string]string
...
+	var specFromProperties = os.Getenv(SpecsDir)
+	if specFromProperties != "" {
+		return specFromProperties
+	}
+	return "specs"
```
nehashri: Here we can use the constant `common. SpecsDirectoryName`. That would better than hard coding the `specs` string here.

vendor/github.com/getgauge/common/common.go
jackperalta: yes - it is intended to have the variable, I used to commented it out to find all usage.
will bring it back in

env/env.go
jackperalta: agree

env/env.go
jackperalta: fine with me

nehashri:
@cla-bot recheck PR.

nehashri:
I had another suggestion, the env variable can take a comma separated list of spec directories. The `GetSpecDir` function can return a array of spec directories. This would be really useful to have.

Also, it looks like this is your first contribution to Gauge. Before we can look at your pull request, you'll need to sign our Contributor License Agreement (CLA).
Please visit https://gauge-cla.herokuapp.com/ to read and sign it.

gaugeci:

jackperalta:
As far as I can see, the current structure of the specification location is not that it supports comma separated lists. Because some places expect to work with one value. But I will see

jackperalta:
Now it would be a difference of passing multiple spec location to Gauge.

When from command line, it would be `gauge run specs1 specs2", so space separation
When from property file, it would be `gauge_specs_dir = specs1,specs2", so comma separation.

Is this acceptable ? @nehashri

jackperalta:
I renamed the existing method `GetSpecFiles` to `GetSpecFilesFrom` (in lack of better name), because I wanted to avoid that all callers have to switch to `[]string`, when not needed. 
And it seems that GO does not support overloading ?!

gaugeci:

util/fileUtils.go
```diff
@@ -152,17 +152,24 @@ func IsDir(path string) bool {
...
+		} else if common.FileExists(path) && IsValidSpecExtension(path) {
+			f, _ := filepath.Abs(path)
+			specFiles = append(specFiles, f)
+		}
 	}
+
 	return specFiles
 }
 
+var GetSpecFilesFrom = func(path string) [] string{
```
nehashri: You are right about go not supporting overloading. Instead of having two functions, In places where only one path is sent, we could convert it to an array before sending it. 
you can convert a string to array as such `[]string{"something"}` .

nehashri:
@jackperalta Also, Before we can look at your pull request, you'll need to sign our Contributor License Agreement (CLA).
Please visit https://gauge-cla.herokuapp.com/ to read and sign it.

util/fileUtils.go
jackperalta: I had this first, but in my eyes this is a bad design, when you force a caller to follow a certain structure. But this of course is easier, when you have a language that supports overloading. 
I will change that.

Regarding the CLA signing, I did this twice yesterday, with getting a big green checkmark when clicking the button - I can do this again, but if this failed the first two times already ?!

zabil: > Regarding the CLA signing, I did this twice yesterday

You don't have to do this again.

nehashri:
@jackperalta Thank you for your contribution :)

Issue #994, opened 2018-04-02
Should not get exception during LSP `autocomplete` call


sguptatw:
**Expected behavior**
Should not get exception during LSP `autocomplete` call

**Actual behavior**
Getting the following exception
```
textDocument/completion: runtime error: index out of range
goroutine 28 [running]:
runtime/debug.Stack(0xc42006d830, 0x9476c0, 0xc9ad00)
	/usr/local/Cellar/go/1.8/libexec/src/runtime/debug/stack.go:24 +0x79
github.com/getgauge/gauge/api/lang.recoverPanic(0xc4203b6000)
	src/github.com/getgauge/gauge/api/lang/server.go:294 +0x57
panic(0x9476c0, 0xc9ad00)
	/usr/local/Cellar/go/1.8/libexec/src/runtime/panic.go:489 +0x2cf
github.com/getgauge/gauge/api/lang.(*files).line(0xca7040, 0xc4204a0060, 0x5a, 0x5, 0x0, 0x0)
	src/github.com/getgauge/gauge/api/lang/file.go:49 +0xec
github.com/getgauge/gauge/api/lang.getLine(0xc4204a0060, 0x5a, 0x5, 0x8fb900, 0xc4204662c0)
	src/github.com/getgauge/gauge/api/lang/file.go:86 +0x4b
github.com/getgauge/gauge/api/lang.completion(0xc4203b6000, 0x17, 0x9eda19, 0x17, 0xffffffffffffff01)
	src/github.com/getgauge/gauge/api/lang/completion.go:57 +0xdf
github.com/getgauge/gauge/api/lang.(*LangHandler).Handle(0xcc4b78, 0xc746e0, 0xc4200164d0, 0xc727e0, 0xc4201e97a0, 0xc4203b6000, 0x0, 0x0, 0x0, 0x0)
	src/github.com/getgauge/gauge/api/lang/server.go:134 +0x24a7
github.com/getgauge/gauge/api/lang.(*LangHandler).handle(0xcc4b78, 0xc746e0, 0xc4200164d0, 0xc4201e97a0, 0xc4203b6000, 0x0, 0x0, 0x0, 0x0)
	src/github.com/getgauge/gauge/api/lang/server.go:86 +0xc3
[...]

PR #997, opened 2018-04-04
handle case where cache is not updated before request processing, #994

riju91:
* goimport commit is not related to the issue


api/lang/file.go
```diff
@@ -46,6 +46,9 @@ func (file *files) remove(uri lsp.DocumentURI) {
 func (file *files) line(uri lsp.DocumentURI, lineNo int) string {
 	file.Lock()
 	defer file.Unlock()
+	if len(file.cache[uri]) <= lineNo {
```
Apoorva-GA: You can use the `getLineCount` method and also check if the file exists.

Issue #1009, opened 2018-04-13
Handle conflicting arguments like failed and repeat


riju91:
### Expected behavior
gauge run --failed --repeat should give invalid command. Same as gauge run --repeat specs

nehashri:
should be available in nightly > 30-4-18

shubhamsc:
should be available in nightly > 06-07-18

PR #1010, opened 2018-04-13
handle argument collision and flag loading for rerun, #1009

riju91:


riju91:
should be available in nightly >= 2018-01-02

PR #1017, opened 2018-04-23
Fixing network connection closed error getgauge/gauge-js#136

Apoorva-GA:
- Changed the way we were handling validation errors when a concept with unimplemented step is used in multiple specs.
- Have done a bit of refactoring in diagnostics.go.

### DIFF
--- a/api/lang/diagnostics.go
+++ b/api/lang/diagnostics.go
@@ -20,6 +20,7 @@ package lang
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sync"
 
 	"github.com/getgauge/common"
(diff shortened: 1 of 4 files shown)


api/lang/diagnostics.go
```diff
@@ -121,20 +144,19 @@ func validateSpecs(conceptDictionary *gauge.ConceptDictionary, diagnostics map[l
 		}
 		createDiagnostics(res, diagnostics)
 		if res.Ok {
-			specs = append(specs, spec)
+			createValidationDiagnostics(validateSpec(spec, conceptDictionary), diagnostics)
```
nehashri: `validateSpec` can return a filtered list of errors here.

Issue #1007, opened 2018-04-11
[Feature Request] Ability to list all tags in a set of spec on gauge cli


vishwas-tw:
### Expected behavior
Tell us what should happen
gauge tags --list

### Gauge version
```
```

zabil:
This can be a documentation plugin that lists the result on the console.
https://manpage.gauge.org/gauge_docs.html

Like spectacle
```
$ gauge docs tags specs
```

jackperalta:
maybe the same solution in the background as wanted here https://github.com/getgauge/gauge/issues/97

zabil:
On second thoughts this must handled by Gauge core as the a plugin to list tags is not discoverable.

The right way is to add a `list` command 
```
$ gauge list --tags
```

That can be expanded to 

```
$ gauge list --specs
$ gauge list --scenarios
```

sriv:
I wonder if having a `search` feature helps an even broader use case?

i.e. 

```
> gauge search foo
[spec] specs/foo.spec
[scenario] specs/foo.spec:3
[tag] specs/foo.spec: 5
```

can be perhaps enhanced to include structured criteria

```
> gauge search tag:foo
[tag] specs/foo.spec: 5
```

and even projections

```
> gauge search foo --select spec,tag
[spec] specs/foo.spec
[tag] specs/foo.spec: 5
```

zabil:
Isn't pipe and grep better?

```
 gauge list --tags | grep term
```
That way the user is in control search filters with full power of grep. 
For windows there's something in powershell

jackperalta:
Oh sorry I guess you mean that people can use the existing grep command...
Anyway I prefer Zabils first suggestion personally

zabil:
> Oh sorry I guess you mean that people can use the existing grep command.

Yes, `grep`, `awk`, `sed` the whole family.

jackperalta:
I'm trying to have a solution for this, but there are still open questions to discuss.

1) Output
What shall the output be ? 
A simple console printout line by line ? 
A well-formatted output somehow organised ?
Like for `--specs` or `--scenarios` - simply the names each in a line ? Etc
Or for tags, shall there be a difference between tags of scenarios and specifications ?
Shall the output be the list of all tags or a set of tags with ignoring multiple occurrences for tags ?

I'm happy to hear your opinions...

2) Warnings
I currently get the following warnings, when running my `gauge list` command

```
WARNING: An illegal reflective access operation has occurred
[...]

PR #1023, opened 2018-04-26
Add functionality to list tags, scenarios,specifications of projects

ajoecker:
Adds a new command to Gauge command line to print out tags, scenarios and specifications of projects.

The command is `list` and has the following options:

- `gauge list --tags specs`, which prints out all tags found in directory `specs`
- `gauge list --scenarios specs`, which prints out all scenarios found in directory `specs`
- `gauge list --specs specs`, which prints out all specifications found in directory `specs`

if the directory is not given, the default one from Gauge is taken.

Print outs are happening line by line and are unique and sorted alphabetically

Addresses feature request: https://github.com/getgauge/gauge/issues/1007


cmd/list.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+var (
+	listCmd = &cobra.Command{
+		Use:     "list [flags] [args]",
+		Short:   "List specifications, scenarios or tags for a gauge project",
+		Long:    `List specifications, scenarios or tags for a gauge project`,
+		Example: `  gauge list --tags specs`,
+		Run: func(cmd *cobra.Command, args []string) {
+			res := validation.ValidateSpecs(getSpecsDir(args), false)
```
nehashri: We don't need to validate all specs here. The `validation.ValidateSpecs` asserts if there are correct implementations in the specific runner. Such errors should not cause this to fail. Also, I don't see the need to start off a language runner here.
We can collect all specs where there are no parse errors and show the respective content.
You can call the parser package's `parser.ParseSpecs` method to get all specifications here.

cmd/list.go
```diff
@@ -0,0 +1,128 @@
...
+		DisableAutoGenTag: true,
+	}
+	tagsFlag      bool
+	specsFlag     bool
+	scenariosFlag bool
+)
+
+func init() {
+	GaugeCmd.AddCommand(listCmd)
+	listCmd.Flags().BoolVarP(&tagsFlag, "tags", "t", false, "List the tags in projects")
```
nehashri: We had a discussion here and we would like to remove the shorthand flags for these commands.
This is mainly because the `--scenarios` flag should have a shorthand of `-s` too. But that would collide with `--specs` flag's shorthand.

cmd/list.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func appendTags(s []string, tags *gauge.Tags) []string {
+	if tags != nil {
+		s = append(s, tags.Values()...)
+	}
+	return s
+}
+
+func uniqueNonEmptyElementsOf(input []string) []string {
```
nehashri: some tests here would be helpful :)

cmd/list.go
jackperalta: Ah, way better - it is way faster now

cmd/list.go
jackperalta: fine with me - changed :)

cmd/list.go
jackperalta: agree - I'm in the process of finding out how tests are written in GO

nehashri:
@ajoecker it looks like you have not signed the CLA with this github user ID. Can you please sign it again.

gaugeci:

ajoecker:
sorry about that - have changed my github id... it is signed now

ajoecker:
I was wondering whether this commands can be extended.

Currently `gauge list --tags specs` or `gauge list --scenarios specs` list all tags/scenarios found in the given spec folder.

I think it would be nice to have something more concrete like `gauge list --tags specs/myspec' to list all tags/scenarios for a given spec (of course this does not make sense for `gauge list --specs`)

Any comments or objections to this ?

zabil:
Yes this makes sense for tags and scenarios.

nehashri:
@ajoecker I will be merging this PR. Can you create a separate issue to track the enhancements.

Issue #144, opened 2015-05-13
Deleting 'specs' folder causes error


prateekbaheti:
_From @timhurkmans on May 13, 2015 14:29_

On gauge-maven-plugin:1.0.1

Create a new folder 'specs2' in the same folder as 'specs'.
Change the specs dir configuration using:

	<specsDir>specs2</specsDir>

Run. Works.
Delete the (now unused) 'specs' folder
Run. Error message:

	[INFO] --- gauge-maven-plugin:1.0.1:execute (default) @ testproject ---
	panic: runtime error: invalid memory address or nil pointer dereference
	[signal 0xc0000005 code=0x0 addr=0x14 pc=0x5097c5]
	
	goroutine 16 [running]:
	github.com/getgauge/gauge/util.func┬À001(0x12a8a690, 0x50, 0x0, 0x0, 0x1b26b8, 0x12afc660, 0x0, 0x0)
	        /tmp/src/github.com/getgauge/gauge/util/finder.go:57 +0x45
	path/filepath.Walk(0x12a8a690, 0x50, 0x12b0df2c, 0x0, 0x0)
	        /usr/local/Cellar/go/1.4.2/libexec/src/path/filepath/path.go:392 +0x97
	github.com/getgauge/gauge/util.FindAllNestedDirs(0x12a8a690, 0x50, 0x0, 0x0, 0x0
	)
	        /tmp/src/github.com/getgauge/gauge/util/finder.go:61 +0xaa
	main.(*specInfoGatherer).watchForFileChanges(0x12a86990)
	        /tmp/src/github.com/getgauge/gauge/spec_details.go:103 +0x35c
	created by main.(*specInfoGatherer).makeListOfAvailableSteps
	        /tmp/src/github.com/getgauge/gauge/spec_details.go:51 +0xe7
	
	goroutine 1 [select]:
	main.(*testRunner).kill(0x12af2060, 0x1b3868, 0x12af2190, 0x0, 0x0)
	        /tmp/src/github.com/getgauge/gauge/runner.go:131 +0x363
	main.executeSpecs(0x7e6600)
	        /tmp/src/github.com/getgauge/gauge/gauge.go:260 +0x1ec
	main.main()
	        /tmp/src/github.com/getgauge/gauge/gauge.go:111 +0x323
	
	goroutine 14 [runnable]:
	net.(*pollDesc).Wait(0x12aa8530, 0x72, 0x0, 0x0)
[...]

PR #1031, opened 2018-05-10
#144 [Statistics] Specification should be skipped if all scenarios are skipped

NivedhaSenthil:


NivedhaSenthil:
Done formatting.

Issue #1067, opened 2018-05-28
CSV table import should have delimiter set as external setting


ajoecker:
### Expected behavior
Parsing a csv file with `;` as delimiter should be successful. In some locales (like german), the excel export to csv generates a csv with `;` as delimiter, as `,` can be part of a cell (e.g. when representing money values like 35,45 EUR)

### Actual behavior
Gauge fails to parse a csv with `;` as delimiter
```
Testing started at 14:13 ...
"C:\Program Files (x86)\Gauge\gauge.exe" run --machine-readable --hide-suggestion --simple-console C:/Users/ajoecker/IdeaProjects/end2end-backend/specs/test.spec:7
[ParseError]C:/Users/ajoecker/IdeaProjects/end2end-backend/specs/test.spec:9 Dynamic parameter <table:data/test.csv> could not be resolved => A step with table <table:data/test.csv>'

Process finished with exit code -1
```
### Steps to reproduce
1. run `gauge init java`
2. generate a csv file in the project with `;` as delimiter
3. use this file in a step
4. execute gauge

### Gauge version
```
$ gauge -v
Gauge version: 0.9.7
Commit Hash: bc9784c

Plugins
-------
flash (0.0.1)
html-report (4.0.4)
java (0.6.7)
screenshot (0.0.1)
spectacle (0.1.2)
xml-report (0.2.0)
```

sguptatw:
needs #1100

sguptatw:
**Question**
Should the delimiter allow multiple characters?

nehashri:
@sguptatw  No we should not have multiple delimiter characters as using multiple delimiters is not ideal.

ajoecker:
I also do not think multiple characters are reasonable. There are (afaik) a minor number of characters, that are used as delimiter in csv (excel does not allow to many different delimiters), all of them are single characters

sguptatw:
[Here](https://community.alteryx.com/t5/Alteryx-Designer-Discussions/Parsing-csv-with-multiple-character-separator/td-p/53214) is a (non-gauge context)discussion where the user wanted multiple characters as a delimiter

That said, its okay if we want to support only single characters. Just wanted to check the expected behavior.

sguptatw:
**Question**
How can I specify something like a space as the separator?
As quoted in [wikipedia](https://en.wikipedia.org/wiki/Comma-separated_values) 
[...]

PR #1068, opened 2018-05-28
1067 cvs delimiter as setting

ajoecker:
Fix for https://github.com/getgauge/gauge/issues/1067

Added a new setting `cvs_delimiter` in `default.properties`, which is per default set to `,` (current behavior)


sriv:
Some unit tests are failing. Appveyor build is a false positive.

ajoecker:
sorry for that... fix incoming

sriv:
merged, this also required an update to the docs: https://github.com/getgauge/docs.gauge.org/blob/master/configuration.rst

Issue #655, opened 2017-05-24
Add an option to --dry-run gauge specs


zabil:
Let the tester do a `--dry-run` to see specification execution order.

```
$ gauge --init java
$ gauge specs --dry-run 
# Specification Heading
  ## Vowel counts in single word	 
  ## Vowel counts in multiple word	 
```

As the name suggests, Gauge must not execute the specifications.

jackperalta:
What shall happen with hooks and or reports ? shall they be triggered and/or created ? Or nothing at all ? Or some of the hooks e.g. ?

nehashri:
Nothing will be triggered. It is a read only operation that tells which specs/scenarios will be executed.

nehashri:
**Question**
- what should be the output if parallel flag is given?
Should the console list all specifications/scenarios per runner, similar to the execution stage?
- Should the dry run flag try and validate all specifications too? (This would mean compiling and bringing up the runner to validate all steps)

sriv:
> what should be the output if parallel flag is given?

Gauge's output (html-report etc) do not differ if it were run in parallel/serial. Going by that fact, I'd say dry run need not change too. 

> Should the dry run flag try and validate all specifications too? (This would mean compiling and bringing up the runner to validate all steps)

This is something we should seek inputs from the community on. Without validation, dry run would be faster, but potentially inaccurate. If the primary usage of dry run is to figure out the applicable specs/scenarios given a filter criteria then I guess gauge need not validate specs. But I feel it would be good to validate, since dry run would then accurately list the scenarios that would get listed.

Side note, if dry run ends up validating, `--validate` becomes redundant.

ajoecker:
Personally I would vote for validating when doing the dry run. I see the dry run as a "what if" situation. What if I really start that.... I would be confused if dry run runs through and then I start and this fails due to validation. I would expect a dry run shows me what would happen if i press the big red button.

MolineroUA:
Hi! I agree with @ajoecker that validation would be a great addition for dry run. Or, as variant, to add possibility to activate the validation from a command
> gauge specs --dry-run 
> gauge specs --dry-run  validate

Thunderforge:
What are the expected durations for the validation portion and the other dry run operations? Are we expecting these to be a matter of seconds?

If both of them are relatively fast, even on large projects, I don't see the downside of pairing them together as part of `--dry-run`. If one or the other is going to take something on the scale of minutes or more, then I could potentially see them being separated.

[...]

PR #1070, opened 2018-05-30
Introducing ' dry-run' flag

ajoecker:
Addressing: https://github.com/getgauge/gauge/issues/655

This pr introduces the `dry-run` flag, so that nothing is executed (or hooks are called) and only the information about what would have been executed is printed.

It is a command line flag and is called via `gauge run --dry-run`.

The implementation is naive, so that I added the flag to the places I expected to suppress something. It is effective, meaning I get the result that I expected, but

a) I do not know whether the approach in general gives meaning
b) something more under the hood is executed, what should not
c) something more under the hood is not executed, what should

Last, I do not know how and where to add unit tests for this - so this first commit does not contain any :(


sriv:
@nehashri/ @ajoecker - is this PR waiting for something? I see that there are some conflicts with `master`.

ajoecker:
its rather old, therefore the conflicts - I could fix that.

nevertheless, as said in my previous comment - it is functional, but I am not sure if it's reasonable and correct.

Also I like to add test cases for this and failed to do so

nehashri:
@ajoecker Thanks for this PR. 
We really need to figure out the correct approach for this. What should be the behaviour of `--dry-run` flag.
- We need to see if validating each and every step is necessary, as this would mean bringing up the runner to do this task. It also raises the question of what level of parsing should be done by Gauge. 
- What happens in case of a parallel run? The specifications are picked up lazily by default in case of a parallel run. So we will not be able to tell which stream runs which specifications unless it is actually run. 

We need to have clarity on this before this feature can be implemented. I am closing this PR and lets discuss on #655 for more details on `--dry-run` flag's behaviour before implementing this.

PR #1091, opened 2018-06-19
Fix broken get started link

gayanhewa:
Existing link is broken - https://gauge.org/get_started.html


zabil:
@gayanhewa thanks for pointing this out.
The original link works now. It was a problem of migrating to our new documentation page.

Issue #1135, opened 2018-08-01
env / default.properties logs_directory setting is ignored


kuechlerm:
### Expected behavior
The logs_directory setting in env/default.properties should be used by gauge. 

### Actual behavior
The logs go into a folder called "logs" no matter what.

### Steps to reproduce
1. create a gauge project
2. change env/default.properties logs_directory
3. gauge run

### Gauge version
```
Gauge version: 1.0.0
Commit Hash: 5a99965

Plugins
-------
csharp (0.10.4)
dotnet (0.1.0)
html-report (4.0.4)
js (2.3.1)
screenshot (0.0.1)
```

ajoecker:
it is getting even more problematic if a later process access the logs_directory environment variable to do things (like a plugin to log also there), this fails then with an exception, as the given directory is not existing

BugDiver:
This issue has been fixed through #1136. Should be available in nightly >= 2-08-2018

PR #1136, opened 2018-08-02
#1135 load logging directory correctly from env variables

ajoecker:
see https://github.com/getgauge/gauge/issues/1135

the environment variables are now loaded before the logger is initialised, so that the logger gets the correct given env varible.


sriv:
cla verified, merging.

Issue #1141, opened 2018-08-08
Proposal add some context of command for before suite


olejkavn:
Proposal to add some commands for gauge to get sush info, it could be somethins like:
gauge get [specs, tags, etc] by [specs, tags]
for me example: gauge get --tags by tagA

Or it can be added in some global test context

Usage of this feature in my case, it's dynamic creation of test run from test:
I deploy feature A and wan't to run test by tagA and create run with my cases in my Test Managment System in @BeforeSuite

Apoorva-GA:
@olejkavn Could you give us an example with the spec file with tags and implementation file.
This will help us understand your use case better.

ajoecker:
personally I see too much interweaving of responsibilities here.

it seems that you like to have Gauge/your implementation doing the complete cycle of controlling your TMS and your automated tests. 
I would strongly advice against that and keep things separate.

Assuming I understood your use case, it is better in my eyes to have a script that will handle the orchestration and creates the run and the triggers Gauge.

olejkavn:
@Apoorva-GA can't understand how it would help you, it 's not depend on specs and their implementations

olejkavn:
@ajoecker Yeap, I understand this. But in case if this stuff would do some script or something else we should support one more mapping, now mapping between TMS case-id and tests(spec files) stored only in gauge tags. We could implement some mechanism for parsing of spec files, but gauge already do it, and some extending of this logic would be nice for us.

ajoecker:
@olejkavn I understand the problem you have, that is way, I created https://github.com/ajoecker/gauge-testrail, my scenarios have a Gauge tag corresponding to TestRail case id.

I am not sure I understand , what you mean with _"should support one more mapping, now mapping between TMS case-id and tests(spec files)"_ - this quite vague. And as said, I would vote against any hard-wiring of gauge and a TMS.

I also think "give me all tags for tag x" is confusing, I guess you mean, give me all case ids of scenarios, which have also tag x in comon ? because a generic "give me all tags", would not return just the caseid, also - in your example - _serviceB_

Gauge has the `list` command (see https://github.com/getgauge/gauge/issues/1007), there was discussion to enhance that to have a more verbose output. This might help you, to retrieve all scenarios/tags for a certain input - but it would still require to parse stuff on your side.

olejkavn:
@ajoecker 

> And as said, I would vote against any hard-wiring of gauge and a TMS.

please check previous comment. there are no any hard-wiring

[...]

PR #1163, opened 2018-09-03
1141 extend list command to filter search by tag

ajoecker:
PR for https://github.com/getgauge/gauge/issues/1141

The idea is to allow the user to filter her search with a given tag. 

The new parameter `taggedWith` is taking a string and would filter the results by presenting only those, that are tagged with the given parameter.

Some examples - given the following spec

```
# A simple spec

## A scenario
tags: tags1, tags2
* Step

## Another scenario
tags: tags3, tags4, tags2
* Step
```
| List Command|Result |Comment |
| ------------- | ------------- |------------- |
| `gauge list --scenarios`  | `A scenario, Another scenario`  | no filtering, behaves as before  |
| `gauge list --scenarios --taggedWith tags3`  | `Another scenario` | Only `Another scenario` is tagged with `tags3`  |
| `gauge list --tags --taggedWith tags1`|`tags1, tags2`|Returns all tags that come together with the given one - the filter itself is returned|
|`gauge list --specs --taggedWith tags2`| |Empty result, as no spec is tagged with `tags2`|


zabil:
Thanks again for this PR
Filtering scenarios by tags is very useful. However it's best to support tag [expressions](https://docs.gauge.org/latest/execution.html#tagged-execution) in the filter. This PR only support filtering with tags.

Can we chain the `--tags` command `--specs` and `--scenario` instead?

For e.g.
```
$ gauge list --specs --tags tag3
[Scenarios]
Another scenario

$ gauge list --specs --tags !tags3
[Scenarios]
A scenario

$ gauge list --specs --tags "tags3 & tags1"
[Scenarios]
A scenario
Another Scenario
```

I am not sure if there's use of parameters when you only use `--tags` maybe it works for validating the expression for e.g

```
$ gauge list --tags
[Tags]
tags1
tags2
tags3
tags4

$ gauge list --tags !tags1
[Tags]
tags2
tags3
tags4

$ gauge list --tags !tags1 & tags5 #tag5 is not in the list use the result to verify
[Tags]
tags2
tags3
tags4

```

ajoecker:
the idea of the PR was to allow filtering based on tags.

Your solution also filters on tags, only that it allows a more complex filtering.

> Can we chain the --tags command --specs and --scenario instead?
Here I have to say that this would confuse me (personally) about what exactly the command would do.

`gauge list --specs --tags tag3` is in my eyes harder to understand as `gauge list --specs --taggedWith tags3`

`gauge list --tags !tags1` would this mean simply to print all tags, except tags1 or all tags, that are not in combination with tags1 ? 

I agree that the syntax for filtering could be extended, so like `gauge list --specs --taggedWith tags1 && !tags2` or so.

But I am not convinced about the propsal with the chaining

zabil:
> But I am not convinced about the proposal with the chaining

My bad, poor choice of words here. Forget that I mentioned chaining.

1) I meant to rename `--taggedWith` to `--tags` to make it consistent with `gauge run --tags` and general convention of not using past tense in command line options
2) Extending the filtering syntax to support whatever is supported by `gauge run --tags`

ajoecker:
ah ok - we are getting closer to agreement here :)

But isn't this then confusing as the `list` command also has the `--tags` parameter for listing tags ?

So to filter scenarios one would use `gauge list --scenarios --tags tags1` and then for filtering tags `gauge list --tags tags1` ? Or `gauge list --tags --tags tags1` (joke)

Isn't it then that `--tags` has multiple meanings ?

zabil:
> But isn't this then confusing as the list command also has the --tags parameter for listing tags ?
Agreed

However with filtering does `--taggedWith` sound ? for e.g
```
$ gauge list --scenarios --taggedWith !tag1
``` 

Reads as list scenarios tagged with not tagged with tag1

How about we do what `git log` does for
```
$ git log ---stat --author <param>
$ git log --graph --after <param>
$ git log --oneline --before
```

For Gauge

```
$ gauge list --scenarios --tag tag1
$ gauge list --scenarios --tag !tag3
$ gauge list --scenarios --tag tag1 & tag3
$ gauge list --tags --tag tag1 # not sure who'd use this so maybe don't allow this
```

ajoecker:
> Reads as list scenarios tagged with not tagged with tag1

For me this is actually _list scenarios tagged with not tag1_

> gauge list --tags --tag tag1 # not sure who'd use this so maybe don't allow this

but this is exactly the use case from https://github.com/getgauge/gauge/issues/1141 (see https://github.com/getgauge/gauge/issues/1141#issuecomment-411725365)

maybe @olejkavn has some input also here

nehashri:
@ajoecker I also like the idea of `--list` here.
Besides this PR has been stale for quite sometime. Can you please update the issue or close it.

ajoecker:
> @ajoecker I also like the idea of `--list` here.
> Besides this PR has been stale for quite sometime. Can you please update the issue or close it.

sorry - i was on holidays.

what do you mean by `the idea of --list` ? 
I would love to continue with that, but as far as I can see, we have not found a common decision yet, no ?

ajoecker:
ok - my suggestion is to scrap this PR, it is already quite old and a lot of discussion is still going on and put the discussion into #1141 ?

Not sure if this is easy to do and agreed on, but I think it is better to discuss on the issue, then an old PR ;-)

PR #1167, opened 2018-09-04
reorganize parser code, perhaps its better now?

sriv:


nehashri:
This looks definitely better than what was there before. Can you please resolve to conflicts here first?

Issue #1157, opened 2018-08-21
Create an npm/yarn installer for Gauge


zabil:
### Current behavior

* Install gauge binaries
* Install node and npm
* `gauge init js` which installs a few more npm packages

JavaScript users prefer npm or yarn taking care of the whole installation. Also in some platforms `npm` is not installed by default causing `gauge init js` to fail.

### Expected behavior

* Install node and npm
* `npm install -g gauge-cli` or `yarn global add gauge-cli` 

This must install the gauge binary and the gauge Javascript plugins and the default [Taiko](https://github.com/getgauge/taiko) template project.

sriv:
`npm install @getgauge/cli` should now install gauge in `node_modules/.bin`.

`-g` should install globally.

sguptatw:
**Question**
With gauge installed from within npm it is not available as an independent command. I mean `npm run test` will run gauge if scripts has `gauge run`. However it is not available as an independent command on the command line. How can a user use gauge commands. Something like `disable telemetry` or check `gauge config` or `gauge help` etc?

sriv:
npm `bin` are installed into `node_modules/bin` which is probably not in your `PATH`. This is standard with npm. You can use [`npx`](https://www.npmjs.com/package/npx) to invoke `gauge` to execute binaries installed via npm.

sguptatw:
The installation of npm and yarn work as expected. However Gauge has issues with local installation. Using package manager the gauge version can be locked to a given project. This will however continue to use the latest of the required plugins.
This may have some issues. 
This is tracked as a part of https://github.com/getgauge/gauge/issues/1184.

PR #1179, opened 2018-09-12
add npm installer for gauge, #1157

sriv:
publishes a npm installer to `@getgauge/cli`

version has been removed from `package.json`. this needs a `npm version <version>` before running `npm publish`

pipeline has been set for this.

### DIFF
--- a/build/npm/.gitignore
+++ b/build/npm/.gitignore
@@ -1,0 +1,3 @@
+
+node_modules
+bin
--- a/build/npm/.gitignore
+++ b/build/npm/.gitignore
@@ -1,3 +1,2 @@
 
 node_modules
-bin
--- a/build/npm/package.json
+++ b/build/npm/package.json
@@ -1,0 +1,35 @@
+{
+  "name": "@getgauge/cli",
+  "description": "npm wrapper for installing gauge command line",
+  "scripts": {
+    "install": "node ./src/index.js",
+    "test": "mocha"
+  },
+  "bin": {
+    "gauge": "./bin/gauge"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/getgauge/gauge.git"
+  },
+  "keywords": [
+    "getgauge",
+    "testing",
+    "gauge"
+  ],
+  "author": "getgauge",
+  "license": "GPL-3.0",
+  "bugs": {
+    "url": "https://github.com/getgauge/gauge/issues"
+  },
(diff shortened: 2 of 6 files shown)


build/npm/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "devDependencies": {
+    "chai": "^4.1.2",
+    "mocha": "^5.2.0",
+    "sinon": "^6.2.0"
+  },
+  "dependencies": {
+    "request": "^2.88.0",
+    "unzipper": "^0.9.3"
+  }
+}
```
BugDiver: Missing `veriosn` field in `package.json` file which is required for installation.

build/npm/src/install.js
```diff
@@ -0,0 +1,70 @@
...
+var getVersion = function(p) {
+    return new Promise( (resolve, reject) => {
+        if (!fs.existsSync(p)) {
+            reject("Unable to find package.json.");
+        }
+        fs.readFile(p, (err, data) => {
+            if(err) {
+                reject(err);
+            }
+            resolve(JSON.parse(data).version);
```
BugDiver: `fs.readFile` gives `data` as `Buffer` as default. Provide encoding as `UTF-8` or use `toString` on the data.

BugDiver: On another thought, we can get rid of reading the file by requireing the `package.json`.

build/npm/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+node_modules
+bin
```
BugDiver: We need the `bin` folder to exists. Why is it ignored?

build/npm/package.json
sriv: see PR [description](https://github.com/getgauge/gauge/pull/1179#issue-214841855).

The `version` is kept out because it needs to be set. This is to ensure that the version is set via the pipeline. Otherwise there needs to be a commit for every deploy.

I am trying to avoid a commit from the pipeline.

build/npm/src/install.js
sriv: ok. will require the `package.json`

build/npm/.gitignore
sriv: the `gauge` inside the bin is a placeholder, to prevent `npm install` from erroring out with "FileNotFound". This folder is not supposed to be tracked by vcs.

build/npm/package.json
BugDiver: Got it.  Make sense.

PR #1199, opened 2018-09-27
Added support to running scenarios with their name

shubhamsc:


filter/specItemFilter_test.go
```diff
@@ -701,3 +701,66 @@ func (s *MySuite) TestToFilterScenariosByUnavailableTags(c *C) {
 
 	c.Assert(len(specs), Equals, 0)
 }
+
+func (s *MySuite) TestToFilterScenariosByName(c *C) {
```
sriv: rename `TestToFilterScenariosByName` -> `TestFilterScenariosByName`

filter/specItemFilter_test.go
```diff
@@ -701,3 +701,66 @@ func (s *MySuite) TestToFilterScenariosByUnavailableTags(c *C) {
...
+
+	var specs []*gauge.Specification
+	specs = append(specs, spec1)
+
+	c.Assert(len(specs[0].Scenarios), Equals, 2)
+	specs = filterSpecsByScenarioName(specs, scenarios)
+	c.Assert(len(specs[0].Scenarios), Equals, 1)
+}
+
+func (s *MySuite) TestToFilterScenarioWhichIsNotExists(c *C) {
```
sriv: rename `TestToFilterScenarioWhichIsNotExists` -> `TestFilterScenarioWhichDoesNotExists`

filter/specItemFilter_test.go
```diff
@@ -701,3 +701,66 @@ func (s *MySuite) TestToFilterScenariosByUnavailableTags(c *C) {
...
+
+	var specs []*gauge.Specification
+	specs = append(specs, spec1)
+
+	c.Assert(len(specs[0].Scenarios), Equals, 2)
+	specs = filterSpecsByScenarioName(specs, scenarios)
+	c.Assert(len(specs), Equals, 0)
+}
+
+func (s *MySuite) TestToFilterMultipleScenariosByName(c *C) {
```
sriv: rename `TestToFilterMultipleScenariosByName` -> `TestFilterMultipleScenariosByName`

gauge/scenario.go
```diff
@@ -103,3 +107,12 @@ func (scenario *Scenario) UsesArgsInSteps(args ...string) bool {
 func (scenario Scenario) Kind() TokenKind {
 	return ScenarioKind
 }
+
+func (scen *Scenario) IsScenarioExists(scenarios []string) bool {
```
sriv: `func (scen *Scenario) IsScenarioExists(scenarios []string) bool {` -> `func (scn *Scenario) HasAnyHeading(headings []string) bool {`

filter/specItemFilter.go
```diff
@@ -71,6 +75,17 @@ func (filter *ScenarioFilterBasedOnTags) Filter(item gauge.Item) bool {
 	return false
 }
 
+func newScenarioFilterBasedOnName(scenariosName []string) *scenarioFilterBasedOnName {
```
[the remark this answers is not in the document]
shubhamsc: there are three filter function for span, tags, and scenarios which are typed based so I create a type base constructor with filter function which will call only when the type is scenarios filter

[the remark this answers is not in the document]
sriv: In that case there are better ways to do this:

- define a type representing `[]*scenario`, and have methods to filter it `ByName`, `ByTag` etc..
- create predicates for the filters and have a generic `Filter` method that will apply the predicate and return only if match is true.

PR #1202, opened 2018-09-28
[Experimental] Add table driven scenario support

sriv:
> NOTE: This is now classified as an experimental feature, and is disabled by default. 

set `allow_scenario_datatable=true` to enable this feature.

What works:

- step dynamic parameters can be resolved using scenario table
- scenarios are run as many times as the table rows
- if scenario does not use the table, it is not repeated, and run once.

TODO:

- `SuiteExecutionResult` has to propogate the table driven scenario results appropriately
- html-report etc should consume and report appropriately.
  - there can be spec as well as scenario table. the user should be able to select rows from either
- refactoring `specExecutor.go`:
  - there are too many branch paths. these can be consolidated, and organized better
- `ProtoScenarioResult` now holds the scenario `DataTable`. So this is duplicate info. there are as many scenario results as the number of row, and each hold a copy of the datatable. Perhaps there is a need for a meta entity - this needs to handle backward compatibility.


osandadeshan:
@sriv 
Can you provide a sample spec file for this feature? I think that will be more helpful to us.

sriv:
@osandadeshan - this is an experimental feature, and is implemented only in gauge core. Reports/IDE plugins do not work well at the moment.

Your spec seems ok, can you try setting `allow_scenario_datatable=true` and run the spec? It should iterate the scenario for every row. 

You can see the console output reporting this well, but html etc reports do not recognize this yet.

osandadeshan:
@sriv I tried this with the following spec.
image
I executed "Login Scenario"

Below is my "default.properties" file
image

When I execute this using IntelliJ IDEA, the html-report is showing an parse error.
```
Errors:
          
            [Parse Error] C:\Users\onimalarathna\IdeaProjects\Gauge-UI-Automation-Demo\specs\test.spec:20 Scenario should have atleast one step => ''
```
          
image

sriv:
@osandadeshan yes, I agree. 

I admit this feature is too rough to be usable now. Consider this as an experiment we did to see how well this could be accomodated. The results look good, but there's still work to be done.

The biggest reason I don't recommend this feature yet, is the lack of support for reporting.

osandadeshan:
@sriv Yes. Actually this feature is very helpful and we'll wait until this implementation and testing phase completed.

mpasha:
@sriv Is this feature available for Windows?
How to set the allow_scenario_datatable, is it in gauge.properties?
It is showing error in the IDE. Kindly help

sriv:
@mpasha https://github.com/getgauge-examples/table-driven-scenario is an example for gauge-python with this feature. you can checkout the project and execute it to see how this feature is used.

Note that IDE and reporting plugins do not support this feature yet.

mpasha:
Hi Srikanth,

 By when this feature will be available in IDE and report plugins.

Regards,
Mehraj Shaik
Devops – Test Automation Engineer

M +65 9237 9395
51B Circular Road, Singapore 049406

sriv:
hi @mpasha - I can't promise any timelines, sorry :(

You can always see what the gauge team's upto here - https://waffle.io/getgauge/gauge

With respect to this feature - honestly we built it as an experiment to see how big a change this will be. 

Making it mainstream would be mostly users such as yourself expressing interest in this. Please raise a new issue (against the plugins that you wish to see this feature work) and we will try to get something out.

We would also appreciate any help (code contribution, testing, docs etc), so feel free to reach out if you are interested!

thanks

PR #1222, opened 2018-10-11
Added warning message for invalid scenarios in running of scenarios with name

shubhamsc:


filter/specItemFilter.go
```diff
@@ -233,11 +233,44 @@ func validateTagExpression(tagExpression string) {
...
+		if isContains(allScenarios,heading) {
+			filteredScenarios = append(filteredScenarios,heading)
+		} else {
+			logger.Warningf(true,"Warning: scenario name - \"%s\" not found",heading)
+		}
+	}
+	return filteredScenarios;
+}
+
+func isContains(scenarios []string, heading string) bool {
```
sriv: `isContains` => `exists`

this can be a local function if it isnt used elsewhere

shubhamsc: it's a local function, start with the small letter. Do you want to rename this?

sriv: I meant something like 

```
var exists = func(headings []string, heading string) bool {
    for _, h := range headings {
        if h == heading {
            return true
        }
    }
    return false
}
```

inside the function `filterValidScenarios`

shubhamsc: it's cleaner than do everything in a single function. I think it's a good way of doing that. Does something wrong with this?

PR #1262, opened 2018-11-30
fixing getting started link to https://gauge.org/getting-started-guid…

yaman:
Link for get started at README.md is https://gauge.org/get-started.html which is wrong. It should be https://gauge.org/getting-started-guide/we-start/


yaman:
thanks for the quick merge @zabil!

zabil:
And thanks for spotting and fixing the error @yaman!

PR #1297, opened 2019-01-08
Ensure that errors from post install propagate to npm

jgwilson42:
We use gauge in our CI environment to control our end to end tests. Unfortunately the npm install sometimes has an error during the download. Because of the lack of return npm does not realise that there has been an error and therefore exits with a 0. This causes our build to continue and to attempt to use gauge when the binary isn't available.

This fix is intended to make the failure to download as part of the install script recognised by npm and therefore exit with the correct status.


sriv:
@jgwilson42 - thank you for the fix.

Issue #1266, opened 2018-12-03
Print version of plugin when it is installed


Thunderforge:
When installing a plugin, you get output like this:

```
>gauge install html-report
........................
Successfully installed plugin 'html-report'.
```
It would be very helpful, especially when looking at server logs, to have the plugin version be included within the printed output. For instance:

```
Successfully installed plugin 'html-report' version 4.0.6.
```

zabil:
@Thunderforge 
You can get this info by running the following command
```
gauge install html-report --log-level debug
```

Please feel free to open the issue and discuss if you had something else in mind.

Thunderforge:
@zabil The main situation that we wanted to include the version was in the case of running Gauge on a VM with no plugins installed. We run `gauge run specs` plus some tag parameters, and then the first thing that Gauge does is install the required plugins (e.g. Java, HTML).

In this use case, adding `--log-level debug` isn't an acceptable workaround because we'll have megabytes of unnecessary logging for our two hour test suite.

It's true that in my original example, it may be useful for `gauge install html-report`. But for the more general example we need, that's not going to work.

Also, it doesn't look like I have the permissions to re-open this issue on GitHub.

sriv:
Gauge attempts to install the plugins specified in `manifest.json` while executing, if it isn't already installed. In this case `--log-level=debug` will apply to execution logs, which is @Thunderforge 's concern.

Reopening this.

Debashis9012:
While installing from the local disk plugin version is not printing.
 
![PluginVersion](https://user-images.githubusercontent.com/46309201/55327817-91457d00-54a8-11e9-84bb-f48b72605693.JPG)

Debashis9012:
This feature has been verified and found fixed with the below build version.

``` gauge-1.0.5.nightly-2019-04-04-darwin.x86_64 ```

PR #1306, opened 2019-01-15
print plugin version

ajoecker:
for #1266

introducing a new variable `Version` in the `InstallationResult` to not misuse the `Info` variable, as this might contain other information.


ajoecker:
will check failing tests

ajoecker:
@nehashri sorry - unfortunately not. I will probably have time by the end of the week

ajoecker:
sorry for the enormous delay... will start to look at it..

ajoecker:
the question for this solution is I assume to print only the version information if one is actually present

so `Failed to install plugin 'unknown' version .` should actually be `Failed to install plugin 'unknown'., right ?

nehashri:
@ajoecker Yes

ajoecker:
> @ajoecker gauge-proto on master seems to be uptodate. any reason why gauge-proto is changed in this PR?

no - no idea why it got updated - not purpose though

nehashri:
You should be able to do it. In your gauge project, goto the gauge-proto directory and checkout the master branch, pull the latest changes. This should remove the proto changes

nehashri:
@ajoecker I will merge this PR and create another commit to fix the gauge-proto issue.

PR #1360, opened 2019-03-06
Sets gauge environment variable based on selected env

### LINKED ISSUE #1333: Set an environment variable with gauge's current environment value
While using Gauge's [environment](https://docs.gauge.org/latest/configuration.html#id1) feature.
There must be a way to access the value of the environment used for execution from the runner.

For e.g

```
$ gauge run specs --env local
``` 

In Javascript
```
console.log(process.env.GAUGE_ENVIRONMENT) // Prints "local"
```

#### WORKAROUND 

Set the environment variable and use it for execution for e.g. in bash

```
$ export GAUGE_ENVIRONMENT=local
$ gauge run specs --env $GAUGE_ENVIRONMENT
```

haroon-sheikh:
Fix for #1333 

This is an initial PR to set `gauge_environment` variable. The language runner could then use this to expose in the API.

### DIFF
--- a/env/env_test.go
+++ b/env/env_test.go
@@ -39,6 +39,7 @@ func (s *MySuite) TestLoadDefaultEnv(c *C) {
 
 	c.Assert(e, Equals, nil)
 	c.Assert(os.Getenv("gauge_reports_dir"), Equals, "reports")
+	c.Assert(os.Getenv("gauge_environment"), Equals, "default")
 	c.Assert(os.Getenv("overwrite_reports"), Equals, "true")
 	c.Assert(os.Getenv("screenshot_on_failure"), Equals, "true")
 	c.Assert(os.Getenv("logs_directory"), Equals, "logs")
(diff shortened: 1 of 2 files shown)


env/env_test.go
```diff
@@ -39,6 +39,7 @@ func (s *MySuite) TestLoadDefaultEnv(c *C) {
 
 	c.Assert(e, Equals, nil)
 	c.Assert(os.Getenv("gauge_reports_dir"), Equals, "reports")
+	c.Assert(os.Getenv("gauge_environment"), Equals, "default")
```
trung: why not using `common.DefaultEnvDir`?

env/env_test.go
```diff
@@ -87,6 +90,7 @@ func (s *MySuite) TestLoadDefaultEnvEvenIfDefaultEnvNotPresent(c *C) {
...
```
trung: why not using `common.DefaultEnvDir`?

env/env_test.go
haroon-sheikh: It's because the test specifically loads the "default" environment. It does not make sense to use the property from common as that can change.

env/env_test.go
haroon-sheikh: Same as before

env/env_test.go
trung: the test name is `TestLoadDefaultEnv`, then it should use `ommon.DefaultEnvDir` constant.

env/env_test.go
trung: this test is confusing, it said `TestLoadDefaultEnvEvenIfDefaultEnvNotPresent` but indeed it loads the "default" env

env/env_test.go
haroon-sheikh: Done.

env/env_test.go
haroon-sheikh: Done.

Issue #1256, opened 2018-11-27
Feature request: auto retry failed test


TzurAnv:
could be useful to have the ability to execute fail scenario automatically at the end of the regular "gauge run" cycle depended on a configurable retry value.
this will help with handling test flakiness.

zabil:
@TzurAnv there a few discussions about flaky tests like https://github.com/getgauge/gauge/issues/1092

Gauge won't be adding this feature. It's best to fix the flakiness.

However, you can use Gauge's has a [rerun failed tests](https://docs.gauge.org/latest/execution.html#re-run-failed-tests) tests features in bash/powershell script after reading [exit codes](https://bencane.com/2014/09/02/understanding-exit-codes-and-how-to-use-them-in-bash-scripts/) from a Gauge run and re-running failed tests till they pass.

Issue #1092, opened 2018-06-19
Feature Request: Automate the 'gauge run -f' possibility


Apnikolak:
### Expected behavior

Currently at the end of suite, when gauge creates the '.gauge/failures.json' file I checked the 'failedItems' and if they exist, save them. Then I get the 'lastRunCmd.json' to get the arguments for the last run command. I am trying to have the test check if the failedItems exist, run 'xyz' spec with the same arguments passed in earlier.

This is similar to how 'gauge run -f' works, where it will re-run only the tests that failed, but I am trying to do this in the project itself, through automation, before it actually generates the html for the pipeline.

Is this currently possible to do? If not could it be a future upgrade? 

### Gauge version
Gauge 0.9.9

zabil:
@Apnikolak is this because you have a flaky test?

Apnikolak:
@zabil Yes, the tests that run overnight, sometimes will fail overnight and when we check them in the morning, they run locally just fine.

So I look through the logs/screenshots to check what might have happened, and sometimes it's a scenario where the step before clicked a drop-down to show more information, and it failed trying to verify the drop-down information, when it runs locally and passes, I'll re-run the stage that the test was on, and it will pass this time around because the drop-down actually displayed this time and it had the info to verify.

We tried to counteract this flaky-ness by adding implicit waits in the methods, so say you are doing anything by Id or by Css Selector, we also have in that method a waitForCssSelectorToBeClickable/Visible or a wait for spinner to be invisible for loads.

zabil:
Closing this issue as the best solution to this is fixing flaky tests.

Apnikolak:
How can 1 fix flaky tests, when they error out to network issues/page loading due to network, etc. 

TestNG has a class file to set re-run attempts. I am trying to manually set this up myself, by -> AfterScenario, get ExecutionContext, check if failing. If true, call a class I named FailedTestConfiguration, in here it opens up cmd, cd to path of project, execute gauge run -f.

This is where I am currently at, but at the moment it send me in an endless loop of rerunning that failed spec.

So if you have any helpful advice on how to set this logic up, i'd greatly appreciate it. @zabil

sriv:
Hi @Apnikolak , you say

> This is where I am currently at, but at the moment it send me in an endless loop of rerunning that failed spec.

This is exactly the reason Gauge cannot help. Gauge can execute specs, and the implementation can determine if the instantaneous state of the application is acceptable or not.

[...]

PR #1370, opened 2019-03-16
Added flag to set max retries count for failed scenario

L2jLiga:
It's kinda like feature request.

Added flag to set max retries for failed scenarios. It could be useful when we have unstable tests which depends on external conditions like internet stability or 3rd-party services.

For example, test may fails because there was bad network conditions and request took too much time.

How it will work:

**Given scenario:**
```
# Simple example

|num|
|  0|
| 10|

## Should retry if fails
* Given <num> equals to 10
```

**When execute:**
 `gauge run specs/ --max-retries-count 2`

**Then output:**
```
# Simple example

   |num|
   |---|
   |0  |
  ## Should retry if fails	 ✘
        Failed Step: Given <num> equals to 10
        Specification: specs/example.spec:8
        Error Message: AssertionError [ERR_ASSERTION]: '0' == '10'
        Stacktrace: 
        AssertionError [ERR_ASSERTION]: '0' == '10'
            at Object.<anonymous> (tests/step_implementation.js:8:10)

   |num|
   |---|
   |0  |
  ## Should retry if fails	 ✘
        Failed Step: Given <num> equals to 10
        Specification: specs/example.spec:8
        Error Message: AssertionError [ERR_ASSERTION]: '0' == '10'
        Stacktrace: 
        AssertionError [ERR_ASSERTION]: '0' == '10'
            at Object.<anonymous> (tests/step_implementation.js:8:10)

   |num|
   |---|
   |10 |
  ## Should retry if fails	 ✔

Successfully generated html-report to => /home/user/test-js/reports/html-report/index.html
Specifications:	1 executed	0 passed	1 failed	0 skipped
Scenarios:	2 executed	1 passed	1 failed	0 skipped
```

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -49,6 +49,7 @@ const (
 	strategyDefault        = "lazy"
 	onlyDefault            = ""
 	groupDefault           = -1
+	maxRetriesCountDefault = 1
 	failSafeDefault        = false
 	skipCommandSaveDefault = false
 
@@ -64,6 +65,7 @@ const (
 	rowsName            = "table-rows"
 	strategyName        = "strategy"
 	groupName           = "group"
+	maxRetriesCountName = "max-retries-count"
 	streamsName         = "n"
 	onlyName            = "only"
 	failSafeName        = "fail-safe"
@@ -111,6 +113,7 @@ var (
 	rows                       string
 	strategy                   string
 	streams                    int
+	maxRetriesCount            int
 	group                      int
 	failSafe                   bool
 	skipCommandSave            bool
@@ -128,6 +131,7 @@ func init() {
 	f.StringVarP(&rows, rowsName, "r", rowsDefault, "Executes the specs and scenarios only for the selected rows. It can be specified by range as 2-4 or as list 2,4")
 	f.BoolVarP(&parallel, parallelName, "p", parallelDefault, "Execute specs in parallel")
 	f.IntVarP(&streams, streamsName, "n", streamsDefault, "Specify number of parallel execution streams")
+	f.IntVarP(&maxRetriesCount, maxRetriesCountName, "c", maxRetriesCountDefault, "Max count of iterations for failed scenario")
 	f.StringVarP(&tagsToFilterForParallelRun, onlyName, "o", onlyDefault, "Specify number of parallel execution streams")
 	f.MarkHidden(onlyName)
 	f.IntVarP(&group, groupName, "g", groupDefault, "Specify which group of specification to execute based on -n flag")
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -50,6 +50,7 @@ const (
 	onlyDefault            = ""
 	groupDefault           = -1
 	maxRetriesCountDefault = 1
+	retryOnlyTagsDefault   = ""
 	failSafeDefault        = false
 	skipCommandSaveDefault = false
 
@@ -66,6 +67,7 @@ const (
 	strategyName        = "strategy"
 	groupName           = "group"
 	maxRetriesCountName = "max-retries-count"
+	retryOnlyTagsName   = "retry-only"
 	streamsName         = "n"
 	onlyName            = "only"
 	failSafeName        = "fail-safe"
@@ -114,6 +116,7 @@ var (
 	strategy                   string
 	streams                    int
 	maxRetriesCount            int
+	retryOnlyTags              string
 	group                      int
 	failSafe                   bool
 	skipCommandSave            bool
@@ -132,6 +135,7 @@ func init() {
 	f.BoolVarP(&parallel, parallelName, "p", parallelDefault, "Execute specs in parallel")
 	f.IntVarP(&streams, streamsName, "n", streamsDefault, "Specify number of parallel execution streams")
 	f.IntVarP(&maxRetriesCount, maxRetriesCountName, "c", maxRetriesCountDefault, "Max count of iterations for failed scenario")
+	f.StringVarP(&retryOnlyTags, retryOnlyTagsName, "", retryOnlyTagsDefault, "Retries the specs and scenarios tagged with given tags")
 	f.StringVarP(&tagsToFilterForParallelRun, onlyName, "o", onlyDefault, "Specify number of parallel execution streams")
 	f.MarkHidden(onlyName)
 	f.IntVarP(&group, groupName, "g", groupDefault, "Specify which group of specification to execute based on -n flag")
@@ -239,13 +243,15 @@ func handleConflictingParams(setFlags *pflag.FlagSet, args []string) error {
 	})
 	if repeat && len(args)+flagDiffCount > 1 {
 		return fmt.Errorf("Invalid Command. Usage: gauge run --repeat")
-
 	}
 	if failed && len(args)+flagDiffCount > 1 {
 		return fmt.Errorf("Invalid Command. Usage: gauge run --failed")
 	}
 	if !parallel && tagsToFilterForParallelRun != "" {
 		return fmt.Errorf("Invalid Command. flag --only can be used only with --parallel")
+	}
+	if maxRetriesCount == 1 && retryOnlyTags != "" {
+		return fmt.Errorf("Invalid Command. flag --retry-only can be used only with --max-retry-count")
 	}
 	return nil
 }
--- a/execution/specExecutor.go
+++ b/execution/specExecutor.go
@@ -274,8 +274,7 @@ func (e *specExecutor) executeScenarios(scenarios []*gauge.Scenario) ([]result.R
 func (e *specExecutor) executeScenario(scenario *gauge.Scenario) (*result.ScenarioResult, error) {
 	var scenarioResult *result.ScenarioResult
 
-	i := 1
-	for {
+	for i := 0; i < MaxRetriesCount; i++ {
 		e.currentExecutionInfo.CurrentScenario = &gauge_messages.ScenarioInfo{
 			Name:     scenario.Heading.Value,
 			Tags:     getTagValue(scenario.Tags),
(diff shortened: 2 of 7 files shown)


nehashri:
@L2jLiga This feature request has been discussed in #1256 and #1092. 
This PR addresses one aspect of retrying when a scenario fails. 
There are a few problems with implementing such a feature (the reason why we think this feature will not add value as discussed in the above issues),
* Retrying all scenarios without knowing the reason for failure will not be helpful. It also masks the actual failure. Failure of a scenario will not always occur due to a environmental conditions, it maybe a genuine failure. Retrying the scenario in this case is useless and is just adding more execution time.
* In reports, how will we show that the scenario had failed before and it has passed after a retry? The reports should show all failures regardless. It should contain the complete execution summary, retry or not, i.e., facts should not be masked in the report. Say even if we find a way to show this on the report, it is just making the report more complex to read. 
* With Gauge and its related projects we are working towards reducing the flakiness of tests. We certainly don't think that masking these failure is the answer, but finding good ways to fix them is the way forward. 

So if you really want to retry scenarios, I would suggest you to use `gauge run -f` feature which will rerun all failed tests. You can also create a wrapper around Gauge and run `gauge run -f` until you have all passing tests. This way the reason for failure is not lost in your reports and reasons are clearly visible.

L2jLiga:
@nehashri first of all we really don't want to have multiple reports that's why I think this options could be part of gauge.

> Retrying all scenarios without knowing the reason for failure will not be helpful.

So.. probably we can add yet another flag, `--retry-only=tags`

> In reports, how will we show that the scenario had failed before and it has passed after a retry?

I guess it's possible to store results of previous retry, for example, in `scenarioResult`, then it can be catched by reporter

> We certainly don't think that masking these failure is the answer, but finding good ways to fix them is the way forward.

I agree with that. Unfortunately we can't make environment 100% stable

nehashri:
> first of all we really don't want to have multiple reports that's why I think this options could be part of gauge.

Having multiple reports may be better than masking issues. 

> So.. probably we can add yet another flag, --retry-only=tags

The problem still persists, we do not know the reason for failure in order to retry. Tags cannot be added after the test has been run

> I guess it's possible to store results of previous retry, for example, in scenarioResult, then it can be catched by reporter

Yes, its definitely possible, but it does not make the reports less complex to read.

L2jLiga:
> Having multiple reports may be better than masking issues.

Not sure that it is masking issues, for me it seems like the way to solve/hide some fluctuating issues with some of tests to make report clearer and to avoid focus on that.

> The problem still persists, we do not know the reason for failure in order to retry.

Only one reason why I propose to have such flag is to define which tests we are marking as unstable. Actually I agree that this is not the way to solve the issue with redundant work

> it does not make the reports less complex to read.

It does not make reports less complex only for those who want to use that feature

zabil:
> I agree with that. Unfortunately we can't make environment 100% stable

I agree to this too. Let's look into merging this and thinking about a few conditions.
If you've run this locally can you share a screenshot of how the HTML report looks with this flag applied?

L2jLiga:
@zabil sure

**command:** gauge run specs/ -c 3

**console output:**

```
# Example spec 1

   |num|
   |---|
   |0  |
  ## Example 1   F
        Failed Step: Given <num> equals to 10
        Specification: specs\example.spec:9
        Error Message: AssertionError [ERR_ASSERTION]: '0' == '10'
        Stacktrace:
        AssertionError [ERR_ASSERTION]: '0' == '10'
            at Object.<anonymous> (tests\step_implementation.js:6:10)

   |num|
   |---|
   |0  |
  ## Example 1   F
        Failed Step: Given <num> equals to 10
        Specification: specs\example.spec:9
        Error Message: AssertionError [ERR_ASSERTION]: '0' == '10'
        Stacktrace:
        AssertionError [ERR_ASSERTION]: '0' == '10'
            at Object.<anonymous> (tests\step_implementation.js:6:10)

   |num|
   |---|
   |0  |
  ## Example 1   F
        Failed Step: Given <num> equals to 10
        Specification: specs\example.spec:9
        Error Message: AssertionError [ERR_ASSERTION]: '0' == '10'
        Stacktrace:
        AssertionError [ERR_ASSERTION]: '0' == '10'
            at Object.<anonymous> (tests\step_implementation.js:6:10)

   |num|
   |---|
   |10 |
  ## Example 1   P

# Example spec 2
  ## Example 2   P

Successfully generated html-report to => C:\Users\Andrey.Chalkin\Downloads\demo\reports\html-report\index.html
Specifications: 2 executed      1 passed        1 failed        0 skipped
Scenarios:      3 executed      2 passed        1 failed        0 skipped

Total time taken: 522ms
```

**Result:**
[html-report.zip](https://github.com/getgauge/gauge/files/2995101/html-report.zip)
image

zabil:
In the sample console output and the HTML reports the scenario and the step fails even if it passes after the three tries. 
The reports must indicate the step passing after three tries and have a way to access the stack trace and screenshot for all the three tries right?

L2jLiga:
At this moment this changes keep only result for last try of each scenario.
Right, in general I would like to see stack trace and screenshot for all retries. Also I think it would be good to indicate somehow such scenarios

L2jLiga:
Given sample was not so good, for example some real task would look like:
```
# Spec

## Scenario   F
...stack trace

## Scenario   F
...stack trace

## Scenario   P

Successfully generated html-report to => C:\Users\Andrey.Chalkin\Downloads\demo\reports\html-report\index.html
Specifications: 1 executed      1 passed        0 failed        0 skipped
Scenarios:      1 executed      1 passed        0 failed        0 skipped


```

And in report you would see that the scenario passed

cmd/run.go
```diff
@@ -211,7 +217,7 @@ func execute(cmd *cobra.Command, args []string) {
 		rerun.WritePrevArgs(os.Args)
 	}
 	installMissingPlugins(installPlugins)
-	exitCode := execution.ExecuteSpecs(specs)
+	exitCode = execution.ExecuteSpecs(specs, maxRetriesCount)
```
nehashri: Instead of passing the maxRetriesCount around, you can add set it in the execution package before. This is the approach we have used here. 
This https://github.com/getgauge/gauge/blob/master/cmd/cmd.go#L177 method inits all package flags before execution.

execution/specExecutor.go
```diff
@@ -272,25 +272,36 @@ func (e *specExecutor) executeScenarios(scenarios []*gauge.Scenario) ([]result.R
...
-	scenarioResult := &result.ScenarioResult{
-		ProtoScenario:             gauge.NewProtoScenario(scenario),
-		ScenarioDataTableRow:      gauge.ConvertToProtoTable(&scenario.ScenarioDataTableRow),
-		ScenarioDataTableRowIndex: scenario.ScenarioDataTableRowIndex,
-		ScenarioDataTable:         gauge.ConvertToProtoTable(&scenario.DataTable.Table),
-	}
-	if err := e.addAllItemsForScenarioExecution(scenario, scenarioResult); err != nil {
-		return nil, err
-	}
+	i := 0
```
nehashri: Some tests need to be added to test this behaviour.

nehashri:
@L2jLiga We had a discussion here and as a first cut we are ok to have only the console report the repeated scenarios. However, please do look at the changes requested before we merge this PR.

nehashri:
> So.. probably we can add yet another flag, `--retry-only=tags`

This needs to also be implemented as a first cut feature. The flag should be able to accept tag expressions as specified in https://docs.gauge.org/latest/execution.html#tag-expressions. 
We may have to follow the same filter logic as https://github.com/getgauge/gauge/blob/master/filter/specsFilter.go#L54.

L2jLiga:
@nehashri thanks you for review. I rebased this PR on the top of master and made suggested changed, could you take a look?

nehashri:
@L2jLiga Your PR looks good. However, we would really like to add the `--retry-only=tags` flag so that scenarios can be marked as flaky. Can you try to add this functionality too?

L2jLiga:
I can try to do it .. In what place is it better to do this filtering?

My guesses are to define this filter inside `executeScenario` or change `Scenario` struct and add field to mark spec as flaky..

nehashri:
@L2jLiga I don't think changing the scenario struct is easy. It would need changes in other parts too. 
The best way is the define this filter inside executeScenario if `max-retries-count` and `--retry-only=tags` are defined. 

But as I mentioned before,
> The flag should be able to accept tag expressions as specified in https://docs.gauge.org/latest/execution.html#tag-expressions. We may have to follow the same filter logic as https://github.com/getgauge/gauge/blob/master/filter/specsFilter.go#L54.

needs to be addressed here

L2jLiga:
@nehashri I tried my best, so seems like I added this flag and it works as expected

execution/specExecutor.go
```diff
@@ -272,25 +273,48 @@ func (e *specExecutor) executeScenarios(scenarios []*gauge.Scenario) ([]result.R
...
-		ScenarioDataTable:         gauge.ConvertToProtoTable(&scenario.DataTable.Table),
-	}
-	if err := e.addAllItemsForScenarioExecution(scenario, scenarioResult); err != nil {
-		return nil, err
+	shouldRetry := RetryOnlyTags == ""
+
+	if !shouldRetry {
+		spec := e.specification
+		tagValues := make([]string, 0)
+		if spec.Tags != nil {
```
nehashri: Tags provided at scenario level should also be considered here. So you may have to look at `scenario.Tags` here.

L2jLiga: Probably, I did not get your point.

As far as I understand `filter.NewScenarioFilterBasedOnTags` accepts `specTags` and returns structure with `Filter` function which can be used to filter scenario (lines 287, 289)

nehashri: Oh, my bad.

Issue #1364, opened 2019-03-07
invalid memory address or nil pointer when running with --tags


AdamJorgenson:
### Expected behavior
Should be able to run specs with tags

### Actual behavior
Cannot run specs with --tags argument

### Steps to reproduce
1.On an OSX 10.13 machine try gauge run --tags TT-Group1 specs/ we also tried gauge run --tags=TT-Group1 specs/ and "gauge run --tags="TT-Group1" specs/
2. If the command  is "gauge run specs/somespec.spec" it works fine

Error ----------------------------------
Panicing : runtime error: invalid memory address or nil pointer dereference
goroutine 1 [running]:
runtime/debug.Stack(0xc0005ab6c0, 0x166e760, 0x1cabc40)
	/usr/local/opt/go/libexec/src/runtime/debug/stack.go:24 +0xa7
main.recoverPanic()
	src/github.com/getgauge/gauge/gauge.go:36 +0x57
panic(0x166e760, 0x1cabc40)
	/usr/local/opt/go/libexec/src/runtime/panic.go:513 +0x1b9
github.com/getgauge/gauge/filter.filterSpecsByTags(0xc000486800, 0x8b7, 0x8b7, 0x7ffeefbffab9, 0x9, 0xc0004a8000, 0x1, 0x1e89000)
	/private/tmp/gauge-20190125-2013-18yzinh/gauge-1.0.4/src/github.com/getgauge/gauge/filter/specItemFilter.go:214 +0xcd
github.com/getgauge/gauge/filter.(*tagsFilter).filter(0xc00fb2ea10, 0xc000486800, 0x8b7, 0x8b7, 0x100dc78, 0x20, 0x1687fe0)
	/private/tmp/gauge-20190125-2013-18yzinh/gauge-1.0.4/src/github.com/getgauge/gauge/filter/specsFilter.go:45 +0xa3
github.com/getgauge/gauge/filter.applyFilters(0xc000486800, 0x8b7, 0x8b7, 0xc0005ab9c0, 0x3, 0x3, 0x900, 0x10000c00019da00, 0x314)
	/private/tmp/gauge-20190125-2013-18yzinh/gauge-1.0.4/src/github.com/getgauge/gauge/filter/filter.go:31 +0x89
github.com/getgauge/gauge/filter.FilterSpecs(0xc000486800, 0x8b7, 0x8b7, 0x3, 0xc0005730a0, 0xc000486800)
[...]

PR #1399, opened 2019-04-26
Lazy parallelization of specification parsing. #1364

BugDiver:


parser/specFileCollection.go
```diff
@@ -0,0 +1,29 @@
+package parser
```
sriv: need licence header

parser/specFileCollection_test.go
```diff
@@ -0,0 +1,21 @@
+package parser
```
sriv: license header

util/rlimitUtil_windows.go
```diff
@@ -0,0 +1,27 @@
...
+
+// Gauge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Gauge.  If not, see <http://www.gnu.org/licenses/>.
+
+// +build windows
```
sriv: let's follow one convention. `_windows` and the `+build` are two different ways of making platform specific implementation. I guess either one is enough

parser/parse.go
```diff
@@ -54,29 +55,56 @@ import (
...
+
 func ParseSpecFiles(specFiles []string, conceptDictionary *gauge.ConceptDictionary, buildErrors *gauge.BuildErrors) ([]*gauge.Specification, []*ParseResult) {
-	parseResultsChan := make(chan *ParseResult, len(specFiles))
-	specsChan := make(chan *gauge.Specification, len(specFiles))
+	sfc := NewSpecFileCollection(specFiles)
+	piChan := make(chan *parseInfo)
+	limit := len(specFiles)
+	rLimit, e := util.RLimit()
+	if e == nil && rLimit < limit {
+		logger.Infof(true, "No of specifcations %d is higher than Max no of open file descriptors %d.\n"+
```
sriv: this could be a debug level log.

parser/specFileCollection.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+type SpecFileCollection struct {
+	mutex     *sync.Mutex
+	index     int
+	specFiles []string
+}
+
+func NewSpecFileCollection(s []string) *SpecFileCollection {
+	return &SpecFileCollection{specFiles: s, index: 0, mutex: &sync.Mutex{}}
```
sriv: the `specFiles` is not mutated as far as I see. What is the purpose of the mutex?

BugDiver: The `specFiles` is not mutated but the `index` is mutated. The `mutex` is there to ensure that two goroutines are not parsing the same file.

sriv: Are `SpecFileCollection` and `NewSpecFileCollection` used outside `parser` package? I could not locate any external usage. If not, then these need not be exported.

PR #1402, opened 2019-05-02
Format the gauge log

shubhamsc:


shubhamsc:
we are using [go-logging](https://github.com/op/go-logging) for logging the gauge logs. and it's used with `logger.info`, `logger.debug`. but we need to set format to show logs as we want.
There are a couple of [option](https://godoc.org/github.com/op/go-logging#NewStringFormatter) in which way we can set logging format and get the expected logs

logger/logger.go
```diff
@@ -53,7 +53,7 @@ const (
 
 var level logging.Level
 var activeLogger *logging.Logger
-var fileLogFormat = logging.MustStringFormatter("%{time:15:04:05.000} %{message}")
+var fileLogFormat = logging.MustStringFormatter("%{time:02-01-2006 15:04:05.000} [Gauge] [%{level}] %{message}")
```
nehashri: `[Gauge]` here is hardcoded, so it may be hard for us to print other plugin id here instead. Is there a way that it can be configured differently?

shubhamsc: [Go-logging](https://godoc.org/github.com/op/go-logging) provides these [options](https://godoc.org/github.com/op/go-logging#NewStringFormatter) to add in Formatter, it doesn't except any outside variable. we can add [Gauge/runner id] with the message which passed through the logger. That need to change all gauge messages which may not efficient way to changes.

nehashri: Merging this for now. We still need to figure out how to remove that hardcoding. The Gauge/Plugin Id should not be decoded and should be set as per need.

PR #1412, opened 2019-05-20
Escape invalid JSON chars while writing logs in JSON format

negiDharmendra:


logger/logger.go
```diff
@@ -173,6 +173,7 @@ func write(stdout bool, msg string, args ...interface{}) {
 		if machineReadable {
 			strs := strings.Split(fmt.Sprintf(msg, args...), "\n")
 			for _, m := range strs {
+				m = escapeInvalidJSONChars(m)
 				fmt.Printf("{\"type\": \"out\", \"message\": \"%s\"}\n", strings.Trim(m, "\n "))
```
nehashri: Instead of creating a json object ourselves, if we use Golang's Json encoder and decoder, it may automatically escape the special characters.

negiDharmendra: Please look at the latest changes.

nehashri:
@negiDharmendra You will have to change https://github.com/getgauge/gauge/blob/master/reporter/jsonConsole.go#L228 too.
Please see if you can reuse the code.

PR #1419, opened 2019-05-22
Add custom writers to intercept Plugin's log

BugDiver:


logger/logger.go
```diff
@@ -76,114 +104,81 @@ func (out *OutMessage) ToJSON() (string, error) {
 }
 
 // Info logs INFO messages. stdout flag indicates if message is to be written to stdout in addition to log.
-func Info(stdout bool, msg string) {
-	Infof(stdout, msg)
+func Info(stdout bool, module string, msg string) {
```
nehashri: The module will be Gauge for most of the use cases over here. 
Can we try avoiding adding this new `module` argument to the function, as it will create confusion. We expect the module to be a blank string for Gauge, for someone looking at this api, it can be confusing. Most of the usecase here will be to use `gauge` as the module.
Instead, why not create a wrapper for these API's? Overloading is not possible, but putting another API, which would accept module should be ok.

BugDiver: Yes. That can be possible, but in that case, all the methods have to be duplicated.
Example: 
```go
func Errorf(stdout bool, module string, msg string, args ...interface{}) {
	write(stdout, msg, args...)
	if !initialized {
		fmt.Fprintf(os.Stderr, msg, args...)
		return
	}
	GetLogger(module).Errorf(msg, args...)
}

// Warning logs WARNING messages. stdout flag indicates if message is to be written to stdout in addition to log.
func Warning(stdout bool, module string, msg string) {
	Warningf(stdout, module, msg)
}

```

Would probably turn into 
```go
func Errorf(stdout bool, msg string, args ...interface{}) {
	write(stdout, msg, args...)
	if !initialized {
		fmt.Fprintf(os.Stderr, msg, args...)
		return
	}
	GetLogger(gauge).Errorf(msg, args...)
}


func Error(stdout bool, msg string) {
	Error(stdout, msg)
}

func ModuleErrorf(stdout bool, module string, msg string, args ...interface{}) {
	write(stdout, msg, args...)
	if !initialized {
		fmt.Fprintf(os.Stderr, msg, args...)
		return
	}
	GetLogger(module).Errorf(msg, args...)
}

func ModuleError(stdout bool, module string, msg string) {
	ModuleError(stdout, module, msg)
}

```
Correct me if I'm wrong?

nehashri: I think that will be better than changing the signature of all the api's. 
- It will reduce confusion while writing logs in other files. 
- Will be more readable.

logger/logWriter.go
```diff
@@ -0,0 +1,77 @@
...
+	case "debug":
+		Debug(w.ShouldWriteToStdout, w.loggerID, m.Message)
+	case "info":
+		Info(w.ShouldWriteToStdout, w.loggerID, m.Message)
+	case "error":
+		Error(w.ShouldWriteToStdout, w.loggerID, m.Message)
+	case "warning":
+		Warning(w.ShouldWriteToStdout, w.loggerID, m.Message)
+	case "fatal":
+		Fatal(w.ShouldWriteToStdout, w.loggerID, m.Message)
```
nehashri: Fatal error on a plugin should not end the Gauge process too. This needs to be handled differently.

logger/logger.go
BugDiver: I think there is another way to achieve this, instead of duplicating all the APIs, we can just reuse the common code for the module writer since it's used at only one place,
Example:  In  the `LogWriter`
```go
case "debug":
    write(w.ShouldWriteToStdout, m.Message)
    if initialized {
        GetLogger(w.loggerID).Warning(m.Message)
    }
case "info":
    write(w.ShouldWriteToStdout, m.Message)
    if initialized {
        GetLogger(w.loggerID).Info(m.Message)
    }
...
...
...
```

logger/logWriter.go
BugDiver: How would you recommend handling it?

BugDiver: We need to aggregate all the fatal errors from the plugins and print it at the end of execution or print them when Gauge's Fataf is used.
Will be tracked as separate issue #1426

logger/logWriter.go
```diff
@@ -0,0 +1,95 @@
...
+			GetLogger(w.loggerID).Error(m.Message)
+		}
+	case "warning":
+		write(w.ShouldWriteToStdout, m.Message)
+		if initialized {
+			GetLogger(w.loggerID).Warning(m.Message)
+		}
+	case "fatal":
+		write(w.ShouldWriteToStdout, m.Message)
+		fmt.Fprintf(os.Stderr, m.Message)
```
nehashri: this may end up writing the log to stderr twice.

PR #1445, opened 2019-07-01
Installed language plugin for validate command if not installed

luciferankon:


cmd/validate.go
```diff
@@ -40,15 +43,21 @@ var (
...
 			}
+			installMissingLanguagePlugin(installPlugins)
 			validation.Validate(args)
 		},
 		DisableAutoGenTag: true,
 	}
 	hideSuggestion bool
 )
 
+func installMissingLanguagePlugin(flag bool) {
```
BugDiver: Can we use the existing `installMissingPlugins` method with addition flag param `languageOnly`? I guess that will remove all the new methods which you have added.

luciferankon: Yeah agreed. But I need the install.LanguagePlugin thing or else I need to pass it around everywhere that might create a mess. Like why AllPlugins should take a flag called languageOnly

BugDiver: > I need to pass it around everywhere

Isn't it better than repeating 3 or more functions?

PR #1492, opened 2019-10-06
Plugin install panic

simross:


simross:
@gaugebot recheck the pull request.

plugin/install/install_test.go
```diff
@@ -94,9 +98,24 @@ func addVersionSupportToInstallDescription(installDescription *installDescriptio
...
+func (s *MySuite) TestReturnsErrorInsteadOfPanicDuringZipFileInstall(c *C) {
+
+	sourceFileName := filepath.Join("_testdata", "plugin.json")
+	targetPathSingleDot := filepath.Join("_testdata", "zip_with_single_dot.zip")
+	targetPathMultipleDots := filepath.Join("_testdata", "zip_with_multiple-dot.s.zip")
+	err := createZipFromFile(sourceFileName, targetPathSingleDot)
+	c.Assert(err, IsNil)
+	defer os.Remove(targetPathSingleDot)
+
+	err = createZipFromFile(sourceFileName, targetPathMultipleDots)
```
negiDharmendra: You can avoid creating zip file in tests.

simross: You're right. But alternative that I see is to create constant archives in `_testdata` directory, which will demand changes to `.gitignore`. If this is more acceptable, I will do so. If you have other solution in mind, please share it with me.

negiDharmendra: If we move the [unzipping](https://github.com/getgauge/gauge/blob/9f8e3e0c8ef323492efba28a191066f7e9d29b04/plugin/install/install.go#L155) after the [platform independent and os compatibility check](https://github.com/getgauge/gauge/blob/9f8e3e0c8ef323492efba28a191066f7e9d29b04/plugin/install/install.go#L159), then we don't need to create the zip file as far as these tests are concerned.

negiDharmendra: Also, the unzipping does not make any sense if the installation is going to fail due to OS compatibility issues. So, it will be better if we unzip the plugin after the [platform independent and os compatibility check](https://github.com/getgauge/gauge/blob/9f8e3e0c8ef323492efba28a191066f7e9d29b04/plugin/install/install.go#L159)

PR #1548, opened 2019-12-26
Using gRPC for documentation plugins.

BugDiver:


cmd/docs.go
```diff
@@ -40,8 +40,11 @@ var docsCmd = &cobra.Command{
 			exit(fmt.Errorf("Missing argument <plugin name>."), cmd.UsageString())
 		}
 		specDirs := getSpecsDir(args[1:])
-		gaugeConnectionHandler := api.Start(specDirs)
-		plugin.GenerateDoc(args[0], specDirs, gaugeConnectionHandler.ConnectionPortNumber())
+		var startAPIFun = func(specDirs []string) int {
```
sriv: rename to `startAPIFunc`

plugin/plugin.go
```diff
@@ -69,7 +72,13 @@ func isProcessRunning(p *plugin) bool {
 }
 
 func (p *plugin) killGrpcProcess() error {
-	m, err := p.ReporterClient.Kill(context.Background(), &gauge_messages.KillProcessRequest{})
+	var m *gauge_messages.Empty
+	var err error
+	if p.ReporterClient != nil {
+		m, err = p.ReporterClient.Kill(context.Background(), &gauge_messages.KillProcessRequest{})
+	} else {
```
sriv: check `p.DocumenterClient` for nil. even though logically it should be ok, it would avoid spurious nil pointer errors cropping up

Issue #1560, opened 2020-01-14
[Gauge] [ERROR] Unable to kill plugin Html Report


Debashis9012:
### Expected behavior
After generating html report, `html-report` plugin should be exited successfully.

### Actual behavior
getting the below error message under gauge.log
`[Gauge] [ERROR] Unable to kill plugin Html Report : rpc error: code = Unavailable desc = transport is closing`

### Steps to reproduce
1. create `gauge-java` project `gauge init gauge-java`
2. execute spec file `gauge run`
3. verify `gauge.log`

####Note: Install gauge, html-report and gauge-java from master.

### Gauge version
```
Gauge version: 1.0.7
Commit Hash: 50fb6484

Plugins
-------
flash (0.0.1)
html-report (4.0.9)
java (0.7.4)
js (2.3.7)
json-report (0.3.2)
python (0.3.8)
ruby (0.5.3)
screenshot (0.0.1)
spectacle (0.1.3)
xml-report (0.2.2)
```

Debashis9012:
Verified and found fixed.

PR #1562, opened 2020-01-16
remove GRPC_STATUS_UNAVAILABLE noise when plugin is shutting down, #1560

sriv:
ref #1560


Debashis9012:
#### Observation:
After generating html report, `html-report` plugin exited successfully without any error.

Tested version:
```
Gauge version: 1.0.7
Commit Hash: 5f1d3011

Plugins
-------
html-report (4.0.9)
java (0.7.4)
```

PR #1582, opened 2020-02-19
Automate release.

BugDiver:


build/make.go
```diff
@@ -450,3 +458,49 @@ func getPackageArchSuffix() string {
...
+	}
+	err = printer.Fprint(fi, fileSet, f)
+	if err != nil {
+		panic(fmt.Sprintf("Unable to write to file %s", file))
+	}
+	err = fi.Close()
+	if err != nil {
+		panic(fmt.Sprintf("Unable to close file %s", file))
+	}
+	runCommand("go", "fmt", file)
```
golangcibot: Error return value of `runCommand` is not checked (from `errcheck`)

.github/workflows/release.yml
```diff
@@ -1,113 +1,307 @@
...
+        run: |
+          cd chocolatey-packages
+          $Env:GAUGE_VERSION | Set-Content version
+          cd gauge
+          ./package.ps1
+          choco push "gauge.$Env:GAUGE_VERSION.nupkg" -k $Env:API_KEY
+
+  brew-release:
+    needs: [github-release]
+    runs-on: macos-latest
```
sriv: this one does not need `macOS` either, we don't seem to invoke `brew` command, and other scripts seem like they can run on linux

build/make.go
```diff
@@ -450,3 +458,51 @@ func getPackageArchSuffix() string {
 	}
 	return "x86_64"
 }
+
+func bumpVersion() {
```
sriv: instead of manipulating the `version.go` using ast parsing, we can externalize the version either as a text file (eg. json) or use env or even use ld flags.

PR #1595, opened 2020-03-03
Fixes report genaration when data driven spec is part of parallel tags filters

### LINKED ISSUE #1584: Report fails to be generated when data driver spec is part of parallel tags filters
### Prerequisites:
**$ gauge -v**
Gauge version: 1.0.7
Commit Hash: ed7b4fd6

Plugins
-------
flash (0.0.2)
html-report (4.0.9)
python (0.3.7)
screenshot (0.0.1)
spectacle (0.1.4)
xml-report (0.2.3)

**default.properties:**
allow_filtered_parallel_execution = true

### Steps:
1. init a python project
2. Create a data driven spec
3. Add a tag to a scenario from the data driver spec (e.g.: no-parallel)
4. Run the specs in parallel and filter the no-parallel scenarios from parallel execution
**$ gauge run specs/ --parallel --only "\!no-parallel"**

### Expected behavior
Specs run successfully and html report is successfully generated

### Actual behavior
Specs run, but the html-report fails to be generate

`Executing in 0 parallel streams.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x85e70d]
goroutine 12 [running]:
github.com/getgauge/html-report/generator.computeTableDrivenStatuses(0xc0002e6000)
        /Users/runner/runners/2.164.0/work/html-report/html-report/generator/transform.go:421 +0x2d
github.com/getgauge/html-report/generator.toSpec(0xc00021d7c0, 0x0)
        /Users/runner/runners/2.164.0/work/html-report/html-report/generator/transform.go:
373 +0x1309
github.com/getgauge/html-report/generator.ToSuiteResult(0xc00002a004, 0x34, 0xc00007ca80, 0x48)
        /Users/runner/runners/2.164.0/work/html-report/html-report/generator/transform.go:78 +0x86a
main.createReport
(0xc0001f1e30, 0x409e01, 0x930280, 0x955c80, 0xc000221500, 0xa513a0)
        /Users/runner/runners/2.164.0/work/html-report/html-report/htmlReport.go:70 +0xc0
main.(*handler).NotifySuiteResult(0xc00000e308, 0xa4cc60
, 0xc0001f1e00, 0xc0001f1e30, 0xc00000e308, 0xc0001f1e00, 0xc000099a80)
        /Users/runner/runners/2.164.0/work/html-report/html-report/handler.go:58 +0x30
github.com/getgauge/html-report/gauge_messages._Reporter_NotifySuiteResult_Handler(0x955c80, 0xc00000e308, 0xa4cc60, 0xc0001f1e00
, 0xc000221560, 0x0, 0xa4cc60`

** See attached a very simple gauge project
[gauge_test_parallel.zip](https://github.com/getgauge/gauge/files/4235723/gauge_test_parallel.zip)

negiDharmendra:
Fixes #1584


gauge/specification.go
```diff
@@ -307,13 +307,15 @@ func (spec *Specification) Filter(filter SpecItemFilter) (*Specification, *Speci
 	specWithOtherItems := new(Specification)
 	*specWithFilteredItems, *specWithOtherItems = *spec, *spec
 	for i := 0; i < len(specWithFilteredItems.Items); i++ {
-		if filter.Filter(specWithFilteredItems.Items[i]) {
+		item := specWithFilteredItems.Items[i]
+		if item.Kind() == ScenarioKind && filter.Filter(item) {
```
sriv: Do we need the check for `item.Kind() == ScenarioKind` - I believe all the three implementation of `SpecItemFilter` have this check inside:

- https://github.com/getgauge/gauge/blob/4df79cc735e766f2fef50fd3bc90e9e5cf4a9445/filter/specItemFilter.go#L51
- https://github.com/getgauge/gauge/blob/4df79cc735e766f2fef50fd3bc90e9e5cf4a9445/filter/specItemFilter.go#L67
- https://github.com/getgauge/gauge/blob/4df79cc735e766f2fef50fd3bc90e9e5cf4a9445/filter/specItemFilter.go#L82

gauge/specification.go
```diff
@@ -307,13 +307,15 @@ func (spec *Specification) Filter(filter SpecItemFilter) (*Specification, *Speci
...
 			specWithFilteredItems.removeItem(i)
 			i--
 		}
 	}
 	for i := 0; i < len(specWithOtherItems.Items); i++ {
-		if !filter.Filter(specWithOtherItems.Items[i]) {
+		item := specWithOtherItems.Items[i]
+		if item.Kind() == ScenarioKind && !filter.Filter(specWithOtherItems.Items[i]) {
```
sriv: same as above

gauge/specification.go
negiDharmendra: Yes, we need to do that as without this we will endup removing non `ScenarioKind` items either from `specWithFilteredItems` or `specWithOtherItems`. 
It was working perior to parallel tag filter fature. As we were not using the filtered  out specs.

negiDharmendra: So probably we can move that check out of those `Filter` function to the function whcih uses them.

sriv: yeah, because right now it looks like we are checking in multiple locations.

PR #1602, opened 2020-03-12
allow spec file extensions to be configured, fixes #1599

### LINKED ISSUE #1599: Feature: Ability to skip execute .md files
We use IntelliJ for java development and we must use gauge for integration testing our REST services.
Yeah, I know IntelliJ plugin is already deprecated. :-/
IntelliJ plugin not support .md files to edit/find/etc gauge specs, but .spec not support eg. plantuml extension.
We start documenting our gauge specs with plantuml in .md files (to easier understand and overview).
But when we execute our specs, gauge find and try execute these .md files thus gauge mark these files it's contains errors because it's not well formatted for gauge.

Can be add new property for default.properties (eg. skip_markdown_files=true/false) to skip markdown files?

Of course we can put these documentations to different directories, but it's to much separation (cloning directory structures, etc.).

sriv:
set `gauge_spec_file_extensions=<comma separated val>` to define valid gauge spec file extensions.

If the above is not set, gauge shall default to `.spec` and `.md` as valid extensions (keeping current behaviour)

### DIFF
--- a/env/env.go
+++ b/env/env.go
@@ -54,7 +54,8 @@ const (
 	allowFilteredParallelExecution = "allow_filtered_parallel_execution"
 	enableMultithreading           = "enable_multithreading"
 	// GaugeScreenshotsDir holds the location of screenshots dir
-	GaugeScreenshotsDir = "gauge_screenshots_dir"
+	GaugeScreenshotsDir     = "gauge_screenshots_dir"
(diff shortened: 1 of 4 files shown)


negiDharmendra:
After setting `gauge_spec_file_extensions` to `.spec` Gauge does not recognise the specs files.
`gaueg run` results in  `No specifications found in specs.`

PR #1608, opened 2020-03-19
Exit with proper status code when specs dir is empty or doesn't exists.

### LINKED ISSUE #1405: Gauge Returns Exit Code 0 When It Fails To Run
### Expected behavior
1. When attempting to run tests ("Gauge run...") in a folder that does not contain a Gauge project file, the Exit Code should be non-zero to denote that the run did not complete successfully.
2. When attempting to run tests in a folder that contains a Gauge project file but you do not point Gauge to a valid spec file or folder containing specs (e.g. "Gauge run adsfasfa" where there's no folder adsfasfa), the Exit Code should be non-zero to denote that the run did not complete successfully.

### Actual behavior
Exit Code 0 is returned.

### Steps to reproduce
1. Open up a command or Powershell window (only tested on Windows) to a directory that does not contain a Gauge project file.
2. Run "Gauge run specs"
3. Check the exit code using either "echo %errorlevel%" (CMD) or "$LASTEXITCODE" (Powershell)
---
1. Open up a command or Powershell window (only tested on Windows) to a directory that does contain a Gauge project file.
2. Run "Gauge run <random name>" where <random name> is a string that does not correspond to a folder in the current directory.
3. Check the exit code using either "echo %errorlevel%" (CMD) or "$LASTEXITCODE" (Powershell)

### Gauge version
Gauge version: 1.0.3
Commit Hash: ff6c0c3

Plugins
-------
csharp (0.10.5)
dotnet (0.1.1)
html-report (4.0.6)
screenshot (0.0.1)

negiDharmendra:
Fixes #1405

### DIFF
--- a/util/fileUtils.go
+++ b/util/fileUtils.go
@@ -18,6 +18,7 @@
 package util
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -80,7 +81,7 @@ func findFilesIn(dirRoot string, isValidFile func(path string) bool, shouldSkip 
 }
 
 // FindSpecFilesIn Finds spec files in the given directory
-func FindSpecFilesIn(dir string) []string {
+var FindSpecFilesIn = func (dir string) []string {
 	return findFilesIn(dir, IsValidSpecExtension, func(path string, f os.FileInfo) bool {
 		return false
 	})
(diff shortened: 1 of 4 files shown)


util/fileUtils.go
```diff
@@ -159,12 +160,23 @@ func IsDir(path string) bool {
 
 // GetSpecFiles returns the list of spec files present at the given path.
 // If the path itself represents a spec file, it returns the same.
+var exitWithMessage = func(message string) {
+	logger.Errorf(true, message)
+	os.Exit(1)
+}
```
BugDiver: Can we use `logger.Fatal` here?

util/fileUtils.go
```diff
@@ -159,12 +160,23 @@ func IsDir(path string) bool {
 
 // GetSpecFiles returns the list of spec files present at the given path.
 // If the path itself represents a spec file, it returns the same.
+var exitWithMessage = func(message string) {
+	logger.Errorf(true, message)
+	os.Exit(1)
```
BugDiver: Can we use `logger.Fatal` here?

negiDharmendra: it prints some extra information 
```
Get Support ----------------------------
	Docs:          https://docs.gauge.org
	Bugs:          https://github.com/getgauge/gauge/issues
	Chat:          https://spectrum.chat/gauge

Your Environment Information -----------
```
I don't think we need this as the error message itself explains the issue.

BugDiver: Ok

PR #1644, opened 2020-05-14
Parallel exec hooks

sriv:
Fixes getgauge/gauge-java#297

Notes:

- With this change - `BeforeSuite` and `AfterSuite` hooks will be executed only once per suite run, provided
  - the runner supports GRPC communication with gauge
  - the runner supports multithreading

GRPC is required because older mode of communication is deprecated, and without multithreading support, gauge would spin of multiple runners, in which case each of the runner instances would have to run the Suite hooks individually.

This brings us to the decision that all runners should support multithreading/multitasking in order to get the Suite Hooks to run as expected.


sethisandeepsethi:
Hi,
Can anyone provide some workaround to this problem in Java.
I want to send an email with test summary after execution of the suite. We are dependent on @AfterSuite method for this.
Currently the email is getting sent that many number of time as we have test cases.

sriv:
@sethisandeepsethi You could do one of the below:

- make your project run in multithreaded mode, install latest gauge-java. This should invoke the `AfterSuite` hook only once per run.
- if the above is not possible, you will have to roll out some cross process synchronization mechanism to indicate that one stream has executed the hook and others can skip it. (see https://stackoverflow.com/a/9577667 for an example)

PR #1649, opened 2020-05-25
Printing console messages via logger package #1542

### LINKED ISSUE #1542: Messages to the console should be printed via logger package
Messages to the console should be printed via logger package

The logger package manages the output to be printed correctly. It prints in json format if machine readable flag is provided. It decides when it is safe to print to the console ( if gauge is started with lsp, it will not print). So it is best to let it handle all prints to process standard  output

There are many places where we directly print to the console in Gauge. These should be printed via the logger.

BugDiver:
Fixes #1542

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -9,7 +9,6 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
-	"os"
 	"path/filepath"
 
 	"github.com/getgauge/gauge/logger"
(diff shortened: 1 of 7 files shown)


cmd/version.go
```diff
@@ -57,6 +57,8 @@ func printJSONVersion() {
 	gaugeVersion := versionJSON{version.FullVersion(), version.CommitHash, make([]*pluginJSON, 0)}
 	allPluginsWithVersion, err := pluginInfo.GetAllInstalledPluginsWithVersion()
 	if err != nil {
+		// logger can not be used, since it breaks the json format,
```
sriv, line 60: I had to fix this for getgauge/gauge-vscode#431. Raised https://github.com/getgauge/gauge/pull/1651

BugDiver: Done

PR #1670, opened 2020-06-04
Strip `v` prefix from brew commit and PR title

dawidd6:
Pull requests and commits should be in the form of: `formula version`.


sriv:
thanks for the patch @dawidd6

Issue #1679, opened 2020-06-19
Environment variable substitution doesn't check properties files


opursail:
### Describe the bug
Properties defined in env/default cannot use substitutions from properties defined an environment (e.g.  env/foo) 

For example: We want to define property templates in default and have values substituted for each environment the specs are executed in  

E.g. 
env/default/global.properties
```
TEMPLATE=something.${MYVAR}
```
env/foo/some.properties
```
MYVAR=testing
```

When you run with --env foo you would expect the property TEMPLATE to resolve to "something.testing"

However you get the error below 
```
'MYVAR' env variable was not set
```

The issue is due to the function below never checking against the properties already loaded from properties files (envVars) and only checking the system environment variables 
```
func substituteEnvVars() error {
	for name, value := range envVars {
		contains, matches := containsEnvVar(value)
		// if value contains an env var E.g. ${foo}
		if contains {
			for _, match := range matches {
				// check if match is from properties file
				// if not, get from system env
				envKey, property := match[0], match[1]
				// error if env property is not found
				if !isPropertySet(property) {
					return fmt.Errorf("'%s' env variable was not set", property)
				}
				// get env var from system
				propertyValue := os.Getenv(property)
				// replace env key with property value
				value = strings.Replace(value, envKey, propertyValue, -1)
			}
			// overwrite the envVar value
			envVars[name] = value
		}
	}
	return nil
}
```

I can submit a PR with tests to fix this issue

### What command(s) did you run when you found the bug?

```
gauge run specs --env foo
```

### Output, stack trace or logs related to the bug 
gauge run --env foo specs
Error ----------------------------------

'MYVAR' env variable was not set

Get Support ----------------------------
        Docs:          https://docs.gauge.org
        Bugs:          https://github.com/getgauge/gauge/issues
        Chat:          https://gitter.im/getgauge/chat

Your Environment Information -----------
        windows, 1.0.6
        go (0.1.2), html-report (4.0.8), java (0.7.2), python (0.3.5), screenshot (0.0.1), spectacle (0.1.3), xml-report (0.2.1)

### Versions 

Issue is present in current code base 

#### Gauge (Output of `gauge -v`)
gauge -v
Gauge version: 1.0.6

Plugins
-------
go (0.1.2)
html-report (4.0.8)
java (0.7.2)
python (0.3.5)
screenshot (0.0.1)
spectacle (0.1.3)
xml-report (0.2.1)

#### Node.js/Java/Python/.Net/Ruby version
NA
[...]

Issue #1428, opened 2019-06-05
Feature request: Reusability for environment variables


Christoffer-Green:
To be able to create a better structure for my defined environments I would like to be able to reuse variables from within the same _.properties_ file.

Example:
```
service_url=http://service.com
api_url=<service_url>/api/endpoint
```

This would enable me to use the inheritance from the **default.properties** file to a much greater extentent. 
It would in some cases also make a large set of env files less fragile e.g you forget to alter a variable and the tests fail because it's trying to access endpoints on a different server.

jbadeau:
Def an interesting feature. :)

zabil:
Gauge uses the standard property files format. Unfortunately the standard property files does not support re-usability. So this can't be implemented in Gauge.

PR #1680, opened 2020-06-23
add substitution of env vars from property files #1679 and #1428

opursail:
If a property can't be substituted from system properties then the ones loaded from properties files are checked

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,96 @@
+name: build
+on:
+  push:
+    branches:
+      - master
+jobs:
+  benchmark:
+    name: Benchmark ${{ matrix.project }} | PARALLEL= ${{ matrix.parallel }}
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    strategy:
+      matrix:
+        project: [java_simple, java_maven, java_gradle]
+        parallel: ['parallel', 'serial', 'multithreaded']
+
+    env:
+      PARALLEL: ${{ matrix.parallel }}
+      BENCHMARK_OUT_DIR: "../out"
+
+    steps:
+      - name: Set up Go 1.13
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
+
+      - name: Setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          path: gauge
+
+      - name: Install Gauge
+        run: |
+          cd gauge
+          go run build/make.go && go run build/make.go --install --prefix=/tmp/gauge --verbose
+          echo "::add-path::/tmp/gauge/bin"
+
+      - name: Install gauge-java
+        run: gauge install java
+
+      - name: Prep benchmark
+        uses: actions/checkout@v2
+        with:
+          repository: getgauge/gauge-benchmarks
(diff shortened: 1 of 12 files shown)


sriv:
thanks for the patch @opursail . I was testing this out, and most of the cases work. I did notice a case where nested access of properties does not work. For ex;

`env/default/default.properties` contains:

```
foo=bar
anotherFoo=${foo}/another
nestedFoo=${anotherFoo}/nested
secondNested=${nestedFoo}/2ndnested
```

resolves to

```
foo: bar
anotherFoo: bar/another
nestedFoo: ${foo}/another/nested
secondNested: ${foo}/another/nested/2ndnested
```

opursail:
> thanks for the patch @opursail . I was testing this out, and most of the cases work. I did notice a case where nested access of properties does not work. For ex;
> 
> `env/default/default.properties` contains:
> 
> ```
> foo=bar
> anotherFoo=${foo}/another
> nestedFoo=${anotherFoo}/nested
> secondNested=${nestedFoo}/2ndnested
> ```
> 
> resolves to
> 
> ```
> foo: bar
> anotherFoo: bar/another
> nestedFoo: ${foo}/another/nested
> secondNested: ${foo}/another/nested/2ndnested
> ```
> 
> So, it looks like we need to figure out a way to allow nested substitution. And this could also lead to handling cyclic references.

@sriv 
I've already got a function to fix nested that deals with cyclic references. But it currently isn't very elegent. So I just went for a simple solution 

Do you want it to deal with nested properties as well?

sriv:
@opursail - I think if this feature is brought in, there will be use cases around nested properties. Unless it is a significantly huge effort, I'd like to address it as well.

What do you say?

> I've already got a function to fix nested that deals with cyclic references. But it currently isn't very elegent.

I think that is ok, we can work on improving it if needed. Nothing is perfect to begin with.

opursail:
Added commit to deal with nested env vars and protect against cyclic references.

sriv:
thanks for the update @opursail, appreciate the effort.

I believe we should be able to report circular dependency a bit better (i.e point out which dependency tree becomes a cyclic graph). 

We [do this for concepts](https://github.com/getgauge/gauge/blob/35a13be9335c919f6eefdcb15ba6688846044de2/parser/conceptParser.go#L326-L365) 

If I extend the logic to properties, I'd imagine something like:

- create a dependency tree of properties.
      - the representation could be a tree structure, (or)
          - create a map of - `prop -> [immediate dependency]`
          - walk the map 
              - if `prop` is found in dependencies, we have a cyclic reference
              - exit when there dependencies are empyy array (i.e we've reached properties with 'pure' values)

This will enable better error reporting - i.e. ideally we'd like gauge to tell the users which properties cause cyclic dependencies, otherwise it could become a nightmare to sort this out!

I know that this is a considerable change, but I think it's best for our users. What do you feel?

opursail:
Added commit to improve detecting and reporting circular dependency as suggested by @sriv

sriv:
Thank you very much for this contribution @opursail. Much appreciated.

I think some of your commits need to be signed - please see https://github.com/getgauge/gauge/pull/1680/checks?check_run_id=884663594

Windows FT has failed, but it seems due to environmental reasons, so I've retriggered. Let's see.

sriv:
@opursail did you mean to close this PR?

opursail:
@sriv - needed to sort out my github signoff config, so have closed this one and re-raised as https://github.com/getgauge/gauge/pull/1700

PR #1686, opened 2020-07-06
benchmark on merge

sriv:


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,102 @@
...
+    needs: [ benchmark ]
+    if: github.event_name == 'pull_request'
+
+    env:
+      FOO: ${{ secrets.FOO }}
+    steps:
+      - name: Test Org Secret
+        run: |
+          echo "FOO: $FOO"
+
```
BugDiver, line 80: Is this needed, or just part of debugging?

sriv: this was part of debugging. fixed.

PR #1700, opened 2020-07-23
Improved environment variable substitution

opursail:
Fix for environment variable substitution doesn't check properties files #1679
Implements solution for #1428
Allows nested property definitions
Prevents cyclic property definitions

Signed-off-by: opursail <15922926+opursail@users.noreply.github.com>


opursail:
@sriv - has the build process for gauge-java changed? This line in the build workflow fails https://github.com/getgauge/gauge/actions/runs/179603855/workflow#L140

sriv:
you are correct, the build process for gauge-java has changed. I've updated the scripts and rebased the PR. Looks like everything is passing now. merging...

Thanks a lot for your contribution @opursail

PR #1710, opened 2020-08-04
Fix error message for lsp server.

BugDiver:
Fixes https://github.com/getgauge/gauge-vscode/issues/439

### DIFF
--- a/api/lang/runner.go
+++ b/api/lang/runner.go
@@ -31,7 +31,7 @@ var lRunner langRunner
 func startRunner() error {
 	var err = connectToRunner()
 	if err != nil {
-		return fmt.Errorf("unable to connect to runner : %s", err.Error())
+		return fmt.Errorf("Gauge could not connect to the runner. Check the logs for more info")
 	}
 	return nil
 }
--- a/api/lang/runner.go
+++ b/api/lang/runner.go
@@ -31,7 +31,7 @@ var lRunner langRunner
 func startRunner() error {
 	var err = connectToRunner()
 	if err != nil {
-		return fmt.Errorf("Gauge could not connect to the runner. Check the logs for more info")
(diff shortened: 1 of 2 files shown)


api/lang/runner.go
```diff
@@ -31,7 +31,7 @@ var lRunner langRunner
 func startRunner() error {
 	var err = connectToRunner()
 	if err != nil {
-		return fmt.Errorf("unable to connect to runner : %s", err.Error())
+		return fmt.Errorf("Gauge could not connect to the runner. Check the logs for more info")
```
zabil, line 34:
```suggestion
		return fmt.Errorf("Gauge could not run the tests. Check the logs for more info")
```

api/lang/server.go
```diff
@@ -278,8 +274,9 @@ func startLsp(logLevel string) (context.Context, *jsonrpc2.Conn) {
 func initializeRunner() error {
 	id, err := getLanguageIdentifier()
 	if err != nil || id == "" {
-		logDebug(nil, "Current runner is not compatible with gauge LSP.")
-		return err
+		e := fmt.Errorf("Current runner is not compatible with gauge LSP. Some of the gauge lsp feature will not work as expected.")
```
zabil, line 277:
```suggestion
		e := fmt.Errorf("There are version incompatibilities between Gauge and it's plugins in this project. Some features will not work as expected.")
```

api/lang/runner.go
BugDiver: This error is not meant for Execution. It's while starting the LSP server. Not sure if it should say something about Execution, could be confusing.

zabil: `the runner` is a bit confusing. Can this be
`Gauge could not initialise. Check the logs for more info`?

BugDiver: 👍

Issue #555, opened 2017-01-24
Incorrect error with missing manifest.json


jgwilson42:
### Expected behavior

If gauge cannot find the manifest.json file then the error should point to this file as being missing.

### Actual behavior

The current error is:
```
Failed to find project directory
```

### Steps to reproduce
1. Install gauge
2. gauge --init java
3. mv manifest.json manifest.json_old
4. gauge specs/

### Gauge version
```
Gauge version: 0.7.0

Plugins
-------
html-report (3.0.1)
java (0.5.1)
```

zabil:
Closing this as an old issue. This may be fixed in the newer releases. 
Will reopen if it still needs fixing.

jgwilson42:
@zabil, please could you reopen this ticket? I am still seeing the issue with the latest version of gauge:

```
Gauge version: 1.0.7
Commit Hash: ed7b4fd6

Plugins
-------
html-report (4.0.9)
java (0.7.4)
screenshot (0.0.1)
```

Steps to reproduce
```
gauge init java
mv manifest.json manifest.json.old
gauge run specs/
```

Output
```
Failed to find project directory
Usage:
  gauge run [flags] [args]

Examples:
  gauge run specs/
  gauge run --tags "login" -s -p specs/

Flags:
  -e, --env string              Specifies the environment to use (default "default")
      --fail-safe               Force return 0 exit code, even in case of failures.
  -f, --failed                  Run only the scenarios failed in previous run. This cannot be used in conjunction with any other argument
<snip>
```

It would be much better if the error specifically mentioned `manifest.json`

PR #1712, opened 2020-08-05
Display the right error message about missing manifest.json

zabil:
Fixes: #555

## Before

```
$ gauge init js_simple
$ mv manifest.json manifest.json.old
$ gauge run specs
Failed to find project directory
Usage:
  gauge run [flags] [args]

Examples:
  gauge run specs/
  gauge run --tags "login" -s -p specs/

Flags:
  -e, --env string              Specifies the environment to use (default "default")
      --fail-safe               Force return 0 exit code, even in case of failures.
  -f, --failed                  Run only the scenarios failed in previous run. This cannot be used in conjunction with any other argument
  -g, --group int               Specify which group of specification to execute based on -n flag (default -1)
  -h, --help                    Help for run
      --hide-suggestion         Hide step implementation stub for every unimplemented step
  -i, --install-plugins         Install All Missing Plugins (default true)
  -c, --max-retries-count int   Max count of iterations for failed scenario (default 1)
  -n, --n int                   Specify number of parallel execution streams (default 8)
  -p, --parallel                Execute specs in parallel
      --repeat                  Repeat last run. This cannot be used in conjunction with any other argument
      --retry-only string       Retries the specs and scenarios tagged with given tags
      --scenario stringArray    Set scenarios for running specs with scenario name
      --simple-console          Removes colouring and simplifies the console output
  -s, --sort                    Run specs in Alphabetical Order
      --strategy eager          Set the parallelization strategy for execution. Possible options are: eager, `lazy` (default "lazy")
  -r, --table-rows string       Executes the specs and scenarios only for the selected rows. It can be specified by range as 2-4 or as list 2,4
  -t, --tags string             Executes the specs and scenarios tagged with given tags
  -v, --verbose                 Enable step level reporting on console, default being scenario level

Global Flags:
  -d, --dir string         Set the working directory for the current command, accepts a path relative to current directory (default ".")
  -l, --log-level string   Set level of logging to debug, info, warning, error or critical (default "info")
  -m, --machine-readable   Prints output in JSON format
```

## After

```
$ gauge init js_simple
$ mv manifest.json manifest.json.old
$ gauge run specs
Failed to find Gauge project directory. Missing manifest.json file.
```

Note: removed the usage instructions as they are not helpful.

Signed-off-by: Zabil Cheriya Maliackal <zabilcm@gmail.com>


vendor/github.com/getgauge/common/common.go
```diff
@@ -70,15 +70,15 @@ type Property struct {
 func GetProjectRoot() (string, error) {
 	pwd, err := os.Getwd()
 	if err != nil {
-		return "", fmt.Errorf("Failed to find project root directory. %s\n", err.Error())
+		return "", fmt.Errorf("Failed to find Gauge project root directory. Missing manifest.json file: %s\n", err.Error())
```
sriv, line 73: this needs to be changed in https://github.com/getgauge/common/common.go and then vendored into gauge again.

Else, the next time this is vendored in, the changes will be lost. It's best to avoid changing the `vendor` directory manually.

zabil:
To be merged after updating `getgauge/common` module pull request https://github.com/getgauge/common/pull/7/files

zabil:
Closing this PR as there are some merge issues. Will raise as separate one with that resolved.

Issue #150, opened 2015-06-04
User PATH env variable gets overwritten on windows installation if PATH is more than 1024 characters


prateekbaheti:
This occurs as mentioned in the [nullsoft installer page](http://nsis.sourceforge.net/Path_Manipulation) if the existing PATH is longer than 1024 characters. 
Need to move to special nullsoft builds which support 8192 string lengths while creating installers.

prateekbaheti:
Fixed from version 0.1.4 onwards for windows installers.

koushik-shetty:
This issue seems to be back in gauge 0.3.0. This install removed all my other path variables. Please Fix the issue.
```
$ gauge -v
Gauge version: 0.3.0

Plugins
-------
html-report (2.0.0)
java (0.3.4)
ruby (0.1.0)
```

apoorvam:
@koushik-shetty We have fixed this issue in the new release v0.3.1. Please download the latest Gauge from [website](http://getgauge.io/get-started/index.html) and try it.

PR #1730, opened 2020-08-31
Set userAgent for downloading from sourceForge

NivedhaSenthil:
Signed-off-by: NivedhaSenthil <nivedhasenthil@gmail.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,7 +46,7 @@ jobs:
         env:
           SIGNING_CERT: ${{ secrets.WINDOWS_SIGNING_CERT }}
         run: |
-          Invoke-WebRequest https://sourceforge.net/projects/nsis/files/NSIS%203/3.06.1/nsis-3.06.1-strlen_8192.zip/download?use_mirror=excellmedia -UseBasicParsing -Outfile ./nsis.zip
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         env:
           SIGNING_CERT: ${{ secrets.WINDOWS_SIGNING_CERT }}
         run: |
-          Invoke-WebRequest https://sourceforge.net/projects/nsis/files/NSIS%203/3.06.1/nsis-3.06.1-strlen_8192.zip/download?use_mirror=excellmedia -UseBasicParsing -Outfile ./nsis.zip
+          Invoke-WebRequest "https://sourceforge.net/projects/nsis/files/NSIS%203/3.06.1/nsis-3.06.1.zip/download" -UseBasicParsing -UserAgent [Microsoft.PowerShell.Commands.PSUserAgent]::Chrome -Outfile "./nsis.zip"
```
BugDiver, line 49: This will cause problems. We need to large strings build of NSIS (strlen 8192)
Refer  https://github.com/getgauge/gauge/issues/150

BugDiver (changes requested):

.github/workflows/release.yml
NivedhaSenthil: Have updated the build uri... dint realise they were different builds

BugDiver (approved):

PR #1733, opened 2020-09-01
Changing the library used to read the properties files

### LINKED ISSUE #1734: Properties are not set as environment variables
### Describe the bug
When there's a large comment in the properties files, the properties after the comment are then skipped.

1. Add this to a property file:

```sh
################################################################
#############################
################################################################
###############################################################
################################################################
##################################################################
##############################################################################
#####################################################
################################################################
##################################################################
################################################################
################################################################
####################################################################
################################################################
#########
#########################################################################
#################
#####################################
test_url=http://testurl
```
2. Attempt to get test_url property. i.e. System.getenv("test_url")

Expected
3. The returned value should be `http://testurl`
Actual:
The returned value is "" (blank)

### What command(s) did you run when you found the bug?

For e.g. 
```
gauge run specs
```

#### Gauge (Output of `gauge -v`)

```sh
Gauge version: 1.1.1
Commit Hash: ea71913

Plugins
-------
flash (0.0.2)
html-report (4.0.12)
java (0.7.9)
json-report (0.3.3)
reportportal (1.7.0)
reportserver (0.2.0)
screenshot (0.0.1)
spectacle (0.1.4)
xml-report (0.2.3)
```

haroon-sheikh:
Signed-off-by: Haroon haroon.sheikh@sky.uk

Problem:
If a properties file has a lot of properties split into different sections by comments. E.g.

```sh
# ------------ Section 1 ------------------
prop=...
..
# ------------ Section 2 ------------------
# property comments
prop2=...
```

Then all of the properties below including prop2 is being skipped by the old properties reader and not set as env vars. 

The dmotylev/goproperties library hasn't had any update since 2014 so it would be a good idea to move onto something like [magiconair/properties](github.com/magiconair/properties) which is more widely used.

Fixes #1734


sriv:
@haroon-sheikh thanks, this is indeed worth changing.

Some builds are failing, let me know if you need any help in sorting these out.

haroon-sheikh:
> @haroon-sheikh thanks, this is indeed worth changing.
> 
> Some builds are failing, let me know if you need any help in sorting these out.

Yes please, I'm not sure why the tests are failing. All of them passing when I run it locally. Is there something environmental I need to do?

haroon-sheikh:
Should be okay now @sriv, can you review for me.

haroon-sheikh:
@sriv I'm not sure why the CodeQL task is failing, are you able to have a look for me please?

sriv:
@haroon-sheikh - I saw a few more commits coming in, just want to make sure if you are happy with this PR now? I'll start my review if you are :)

haroon-sheikh:
> @haroon-sheikh - I saw a few more commits coming in, just want to make sure if you are happy with this PR now? I'll start my review if you are :)

Yep, I'm happy for you to review please.

sriv:
Test circular ref:

`env/default/default.properties` :

```
...
circular = ${anotherCircular}
anotherCircular = ${circular}
...
```

Gauge 1.1.3 shows:

```
...
[Gauge]
circular reference found in env variable 'circular'. Check for cyclic property definitions in:
circular=${anotherCircular}
anotherCircular=${circular}
...
```

With this PR:

```
[Gauge]
Failed to load env. Failed to parse: /tmp/foo/env/default/default.properties. circular reference
```

I feel the older error message is better. I remember there was conscious effort put in by the contributor (@opursail ) to bring out exactly where the circular reference comes from.

This check is also recursive, so it finds out circular references even when nested. Ex:

```
circular = ${anotherCircular}
anotherCircular = ${thirdCircular}
thirdCircular = ${circular}

```

gives:

```
[Gauge]
circular reference found in env variable 'circular'. Check for cyclic property definitions in:
circular=${anotherCircular}
anotherCircular=${thirdCircular}
thirdCircular=${circular}
```

haroon-sheikh:
Thanks for reviewing @sriv.

The circular error message is from the library. I thought it would be better to use that than to implement our own.

This is still circular not recursive:

```sh
circular = ${anotherCircular}
anotherCircular = ${thirdCircular}
thirdCircular = ${circular}
```

${circular} depends on ${anotherCircular} depends on ${thirdCircular} depends on ${circular}.

sriv:
Yes, I figured that. But if you see it from a user's perspective, the experience is a bit broken. I am insisting on this, because at least with gauge, you can set properties in a file as well as env (and there is precedence). It may not always be possible to isolate the cause.

We have two ways forward:
- send a PR upstream to improve the messaging
- patch it up in gauge

The former is better, latter is possibly simpler. I'll let you choose :P

sriv:
Test circular reference in another file:

Say we have two properties files, and a circular reference chain across the two, ex:

`env/default/default.properties`:

```
circular = ${anotherCircular}

anotherCircular = ${thirdCircular}
```

and `env/default/another.properties` :

```
thirdCircular = ${circular}
```

Gauge 1.1.3 shows:

```
[Gauge]
circular reference found in env variable 'thirdCircular'. Check for cyclic property definitions in:
thirdCircular=${circular}
circular=${anotherCircular}
anotherCircular=${thirdCircular}
```
with this change:

```
No errors found.
```

haroon-sheikh:
I've now created a PR to include the circular ref key in the error message https://github.com/magiconair/properties/pull/49

haroon-sheikh:
In terms of the circular reference in properties. 
The reason its not erroring is because when it's trying to resolve the properties, it also does a os.Getenv(key), which by default returns an empty string if the env var does not exist?

Do we now want to consider properties with empty values invalid?

sriv:
> The reason its not erroring is because when it's trying to resolve the properties

Can you give an example?

All my tests so far have been comparing gauge's behaviour before and after this change. Of course, there can be pre-existing issues.

haroon-sheikh:
> > The reason its not erroring is because when it's trying to resolve the properties
> 
> Can you give an example?
> 
> All my tests so far have been comparing gauge's behaviour before and after this change. Of course, there can be pre-existing issues.

I've now fixed and added a unit test for the circular ref across multiple envs. Just need to wait for the error message PR to be merged.

sriv:
I see that https://github.com/magiconair/properties/pull/49 is merged, let me know when I can take another look.

Also, I haven't looked at multiple env yet, but thanks for addressing it.

My earlier test was : In a single environment, have properties with circular references, but the properties are spread across files.

haroon-sheikh:
@sriv I've now updated the branch with latest version of magiconair/properties. Could you please have another look now?

sriv:
hi @haroon-sheikh - the change makes it better, but I still feel that `master` has more details in the error message.

ex: with 

```
foo = ${bar}
bar = ${foo}
```

`master` shows:
```
circular reference found in env variable 'foo'. Check for cyclic property definitions in:
foo=${bar}
bar=${foo}
```

whereas this change shows:
```
Failed to load env. Failed to parse: /tmp/foo/env/default/default.properties. circular reference in "bar = ${bar}"
```

The cyclic graph is missing, and I worry that it can be hard to locate it when it happens.

haroon-sheikh:
@sriv I'm now thinking if we should actually do this update. I was looking to implement a simpler solution and remove custom code but it looks like with our requirements, it's required to implement our own version of properties module. Shall we leave it as is for now?

sriv:
I think we should address the fact that the current dependency is not maintained. And I appreciate all the work that has gone into this PR. 

I'll take a shot at looking at some way to get the cyclic graph in. Since this fixes #1734 (which I believe is more important than the cyclic deps), I'd like this to be pursued (and am happy to help get this in).

All this deliberation is just because this is sort of regressing :(

haroon-sheikh:
Yes please, I'm happy for you to takeover from here to fix the regression if you've got some time.

haroon-sheikh:
I've pushed another failing test if you can have a look at that also please.

sriv:
hi @haroon-sheikh I finally got to spend some more time on this. I felt the best place for this change would still be in the upstream, so I've raised another pull request: https://github.com/magiconair/properties/pull/50

sriv:
@haroon-sheikh https://github.com/magiconair/properties/pull/50 is now merged and released. Can you please try updating the dep?

Also, note instead of `properties.LoadFile` gauge could probably benefit better with [`properties.MustLoadFiles`](https://pkg.go.dev/github.com/magiconair/properties#MustLoadFiles) which could load all the properties files in order. This would mean more refactoring, so if you feel we could take this up as a separate task, I can understand.

haroon-sheikh:
Thanks @sriv. I've now updated the lib and failing unit tests. There's a unit test scenario `TestLoadDefaultEnvWithMissingSubstitutedVariableAcrossMultipleFiles` which I'm not able to fix, would you be able to give me a hand with this?

sriv:
Thanks @haroon-sheikh . @BugDiver has accepted to look into this further.

BugDiver:
@haroon-sheikh I looked into the PR. The failing test is because of the way we are substituting the env. We parse the env `${a}` and set `a` as key and `${a}` as it's the default value which causes the circular ref, but we need to do it because it can refer to other envs as well.

To fix it I tried using `MustLoadFiles` which loads all the files (we do not have to loop over files as @sriv was suggesting),. This makes the test pass, and the output is also right. The drawback is this function panic on errors, and the error message contains timestamps (because of panic I guess).

IMO library method should return an error instead of panicking. I trying to find a way if I can recover from panic and create a proper error message.

haroon-sheikh:
Thanks for looking into this @BugDiver. I think one of the problems in the new lib is that the files take priority over env vars. And we want env vars to take precedence instead. Do you think it's worth us creating our own properties module in the project (or is it an overkill)? That would let us define the logic in the way we want.

BugDiver:
>  I think one of the problems in the new lib is that the files take priority over env vars.

True

> Do you think it's worth us creating our own properties module in the project (or is it an overkill)?

Not sure if it will be overkill. But I think it's quite an effort

Also since this lib panic on errors, it's hard for us to give proper error messages.

We can either submit patches to the lib or fork it and change it according to our requirements.
If we want to maintain our own, I guess we could just fork the older one or the new one and make changes.

BugDiver:
@haroon-sheikh I have pushed a few changes to the `env` branch. Please have a look at it, try to install `gauge` from this branch, and check if all the cases are covered and it's giving the desired output/error. Also Please check the tests as well

haroon-sheikh:
Thanks @BugDiver. I've had a look at the env branch and it seems to do the job and all of the tests are passing too.

BugDiver:
@haroon-sheikh Thanks for the quick response. If you want you can rebase your PR with the branch, or I can raise a PR from that branch

haroon-sheikh:
Please raise another PR and I'll close this one down. ;)

Issue #1244, opened 2018-11-14
Update notice.md to include all dependency lisences.


nehashri:
The notice.md file is quite old and not maintained. This needs to be updated to reflect the latest dependency and changes.
The skel file also includes a notice variable, which is used to generate the notice file while installing Gauge for the first time. This needs to be updated too. It would be ideal if the content for this is copied from the notice.md file itself instead of maintaining it in two places.
Gauge also never updates this file after it is created for the first time. This needs to be changed and it should be updated every time a Gauge update is made to reflect the latest notice.

In summary, things to be done
- Update notice.md file with all latest dependency licenses.
- Generate the notice file for first time installation from notice.md file instead of the variable currently used.
- Every update of Gauge should also update the notice.md file created.

PR #1747, opened 2020-09-08
Update the notice list

zabil:
Fixes: #1244

Switch to plain text format for the notice
To generate the notice install

https://github.com/google/go-licenses

and run

$ go-licenses csv "github.com/getgauge/gauge" > notice.txt

Signed-off-by: Zabil Cheriya Maliackal <zabilcm@gmail.com>


sriv:
should this be automated? i.e. have an action to run and checkin if there is a diff?

zabil:
Might be an overkill to automate this especially when adding libraries are rare. I recommend running this command when a library is added and making it a part of the review process instead.

sriv (approved):

notice.txt
```diff
@@ -0,0 +1,23 @@
+github.com/getgauge/gauge,Apache-2.0
```
sriv, line 1: This is circular, i.e. gauge depending on gauge.

sriv (approved):

Issue #1746, opened 2020-09-08
file variable cannot be passed in Table driven scenario


z04jmath:
Using 
```
Gauge version: 1.1.3
Commit Hash: 9d3a3d5

Plugins
-------
html-report (4.0.10)
js (2.3.10)
json-report (0.3.3)
screenshot (0.0.1)
xml-report (0.2.3)
```

When using table driven scenario like this

```md
## temp
| case | data |
| - | - |
| 1 | <file:Data/json/1.json> |
| 2 | <file:Data/json/2.json> |
* test <data>
```

Content of file is not putted into data variable, it contains `file:Data/json/1.json` instead of its content

sriv:
Yes, currently table driven scenarios is an experimental feature, and has limitations. This is one of those limitations and needs to be addressed.

BugDiver:
Fixed via #1765

PR #1765, opened 2020-09-23
Resolving special params in scenario data table #1746

BugDiver:
Signed-off-by: BugDiver <vinaysh@thoughtworks.com>


execution/specExecutor.go
```diff
@@ -339,6 +338,7 @@ func (e *specExecutor) addAllItemsForScenarioExecution(scenario *gauge.Scenario,
 		return err
 	}
 	if scenario.ScenarioDataTableRow.IsInitialized() {
+		parser.GetResolvedDataTablerows(scenario.ScenarioDataTableRow)
```
sriv, line 341: can we add some tests for this?

sriv (changes requested):

PR #1770, opened 2020-09-30
Fix broken user doco link on readme

craigvl:
Current link to user doco is broken on readme


sriv (approved):

sriv:
Thank you for sending the correction @craigvl !

PR #1794, opened 2020-11-27
use env files for github actions

sriv:


zabil:
golangci-lint / lint is failing, maybe worth fixing those issues as a part of this PR too?

sriv:
yes, am looking at those. I had to update the gh action for golang-ci lint and as part of that I updated the linter version, I guess new issues are reported (which is good).

fixing those.

zabil (approved):

PR #1860, opened 2021-01-15
Fixes an issue with env variable is to empty string

haroon-sheikh:
Following this change https://github.com/getgauge/gauge/commit/4151d5b48c3973d5d569c565874fa715ce2bf86e

We no longer ignore env var if its set to an empty string.

This change is allow the value from properties file when env var value is an empty string


sriv (approved):

sriv:
thanks @haroon-sheikh

PR #2037, opened 2021-07-06
add arm64 plugin support

sriv:
Plugins, so far, are either platform independent or only support `x86` and `x86_64` arch.

This PR adds plugins targetting `arm64` to be installed.

Signed-off-by: sriv <srikanth.ddit@gmail.com>


xemle (approved):

xemle (approved):
Thank you for your improvement

PR #2045, opened 2021-07-14
allow x86_64 binaries on arm64 for darwin

sriv:


plugin/install/install.go
```diff
@@ -139,15 +139,20 @@ func isPlatformIndependent(zipfile string) bool {
 	return !re.MatchString(zipfile)
 }
 
-func isOsOSCompatible(zipfile string) bool {
+func isOSCompatible(zipfile string) bool {
 	os := runtime.GOOS
 	arch := getGoArch()
+	fmt.Println(os + "/" + arch)
```
zabil, line 145: Is do we really want to print this?

sriv: nope, my bad. fixed.

zabil (approved):

PR #2047, opened 2021-07-15
introduce gauge_data_dir

### LINKED ISSUE #2046: Support for CSV file per environment.
## Problem:
We use CSV files for Test Data extensively in our Gauge Tests. Its a very nice feature, especially when you can loop thru the various rows and run many different situations using only a few test scenarios defined in the spec files. 

The Issue is that CSV files are hard coded in the spec files, and this is a source of a lot of issues. We have many test environments (dev, test, staging, pre-prod, uat). Unfortunately, the data is not the same in each environment. 

One CSV is not suitable for all the environments. This leaves us with having multiple test specs for each environments.
When you have 1000s of tests, multiplied by the number of environments, it causes a lot of unnecessary duplication and maintenance. 

It also defeats the environments feature of Gauge, which I believe is one of the most important feature of the framework. 

This is not a new issue. It has been requested before; and unfortunately ignored. This is a fundamental issue. 
I think this issue needs to be given importance as this issue has come up many times before. We ask our testers to develop tests that are environment agnostic and data driven, but then we cannot support their effort because of this flaw. Test data is a major aspect of the environments feature and it needs to be supported. 

I have also implemented my own JSON data source which can be used in tests. However, I cannot offer the looping functionality (of rows) that I get with the CSV data source. I cannot leverage the tagging functionality either, because the data per environment is different. I cannot say run all tests tagged with `sanity` tag along with specifying the environment `--env staging` and have the test pick up the correct data. CSV file reference is hard coded.

Currently, we are unable to use the environments functionality as we would like and have had to duplicate our tests in order to test on our various environments. I know this feature presents a technical challenge for the team, but it is also a big issue that needs a solution. 

I use Gauge with Java, Maven and Spring. I have not come across any reasonable work around to this issue. 

## The Ask:
Provide an Environments ready Test data source that allows us to loop thru a set of data. 

We define one test scenario, give it a data source of 100 records and press play. You have 100 Tests completed. Add additional data sources to account for environment changes and you are golden. CSV data source matches most of those requirements with the exception of the hard coded path to the csv file. 

## Possible Solution(s):

* Add the ability to add a reference to the environment in the csv file path.
* * `table: /env/${env}/data/mydata.csv`
* Add Filtering functionality on the CSV file so we can determine which rows to use for a scenario.
* * Table-rows functionality that already exists isn't all the useful. It relies on knowing the exact location of your data in every csv file and cannot be used when running multiple scenarios/spec files. 
* * Having a filtering option will check a row against a condition and either include the row or exclude it.
* * `table: /env/default/data/mydata.csv:env==staging`
* * Here the `env` is a column header inside the CSV file and `staging` is the value we are looking for in the column to include the data in the test.
* Add support for Excel file `.xlsx` as a data source. Gauge can use worksheet name to pick up the correct test data per environment.
* * `table: /env/default/mydata.xlsx`
* * create worksheet with name `staging` and add all data related to staging environment in there.
* * loop thru all the data on the `staging` worksheet. 
* Add the ability to loop the scenario on a list or set of data on demand.
* * Allow us to implement some sort of interface that will return a list of records that gauge can then loop thru.
```
interface TestDataSource
{
    List getTestData(ScenarioName)
}
```

Please Find a solution to this issue. We would really appreciate it. 

```
Gauge version: 1.2.1
Commit Hash: fa7918b

Plugins
-------
html-report (4.0.12)
java (0.7.15)
json-report (0.3.5)
screenshot (0.0.1)
```

sriv:
allow csv/txt data files to be in subdirectory, configured by gauge_data_dir

fixes #2046

Signed-off-by: sriv <srikanth.ddit@gmail.com>

### DIFF
--- a/util/fileUtils.go
+++ b/util/fileUtils.go
@@ -228,7 +228,7 @@ func GetPathToFile(path string) string {
 	if filepath.IsAbs(path) {
 		return path
 	}
-	return filepath.Join(config.ProjectRoot, path)
+	return filepath.Join(config.ProjectRoot, env.GaugeDataDir(), path)
 }
 
 // Remove removes all the files and directories recursively for the given path
--- a/util/fileUtils.go
+++ b/util/fileUtils.go
@@ -228,7 +228,15 @@ func GetPathToFile(path string) string {
 	if filepath.IsAbs(path) {
 		return path
 	}
-	return filepath.Join(config.ProjectRoot, env.GaugeDataDir(), path)
+
+	gaugeDataDir := env.GaugeDataDir()
+	if gaugeDataDir != "." && filepath.IsAbs((env.GaugeDataDir())) {
+		logger.Warningf(true, "'gauge_data_dir' property should be relative to Project Root. Found absolute path: %s", gaugeDataDir)
+	}
+
+	pathToFile := filepath.Join(config.ProjectRoot, gaugeDataDir, path)
+	logger.Debugf(true, "Reading data file: %s", pathToFile)
+	return pathToFile
 }
 
 // Remove removes all the files and directories recursively for the given path
--- a/util/fileUtils.go
+++ b/util/fileUtils.go
@@ -231,7 +231,7 @@ func GetPathToFile(path string) string {
 
 	gaugeDataDir := env.GaugeDataDir()
 	if gaugeDataDir != "." && filepath.IsAbs((env.GaugeDataDir())) {
-		logger.Warningf(true, "'gauge_data_dir' property should be relative to Project Root. Found absolute path: %s", gaugeDataDir)
+		logger.Warningf(true, "'gauge_data_dir' property must be relative to Project Root. Found absolute path: %s", gaugeDataDir)
 	}
 
 	pathToFile := filepath.Join(config.ProjectRoot, gaugeDataDir, path)
--- a/env/env.go
+++ b/env/env.go
@@ -46,6 +46,7 @@ const (
 	// GaugeScreenshotsDir holds the location of screenshots dir
 	GaugeScreenshotsDir     = "gauge_screenshots_dir"
 	gaugeSpecFileExtensions = "gauge_spec_file_extensions"
+	gaugeDataDir            = "gauge_data_dir"
 )
 
 var envVars map[string]string
@@ -315,4 +316,13 @@ var GaugeSpecFileExtensions = func() []string {
 // AllowCaseSensitiveTags determines if the casing is ignored in tags filtering
 var AllowCaseSensitiveTags = func() bool {
 	return convertToBool(allowCaseSensitiveTags, false)
+}
+
+// GaugeDataDir gets the data files location. This location should be relative to GAUGE_PROJECT_ROOT
+var GaugeDataDir = func() string {
+	d := os.Getenv(gaugeDataDir)
+	if d == "" {
+		return "."
+	}
+	return d
 }
--- a/parser/convert.go
+++ b/parser/convert.go
@@ -140,7 +140,12 @@ func (parser *SpecParser) initializeConverters() []func(*Token, *int, *gauge.Spe
 	}, func(token *Token, spec *gauge.Specification, state *int) ParseResult {
 		resolvedArg, err := newSpecialTypeResolver().resolve(token.Value)
 		if resolvedArg == nil || err != nil {
-			e := ParseError{FileName: spec.FileName, LineNo: token.LineNo, LineText: token.LineText(), Message: fmt.Sprintf("Could not resolve table from %s", token.LineText())}
+			errMessage := fmt.Sprintf("Could not resolve table. %s", err)
+			gaugeDataDir := env.GaugeDataDir()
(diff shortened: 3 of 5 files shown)


sriv:
Try running this build against this sample: [2046_sample_env_specific_data.zip](https://github.com/getgauge/gauge/files/6825358/2046_sample_env_specific_data.zip)

alpha1592:
I tried out the sample project and the build 1.3.3.
Its looking good. I like it.

Couple of points that I noticed while playing around with this.
* Gauge should output the path to the table being used during execution on the console. 
* * This would help when troubleshooting issues.
* * I used the default log level, so not sure if this info is logged on debug or trace levels.
* * I think it should be logged on info level so at the start of test, user is clear where the data is coming from
* * Similarly, i think the path to the table should be logged in the html/other reports
* When Gauge is unable to find the Table, the error is not clear.
* * I provided an invalid path to `gauge_data_dir`
* * `gauge_data_dir = /env/bar/data2` instead of the correct `gauge_data_dir = /env/bar/data`
* * Error: `Could not resolve table from table: bar.csv => 'table: bar.csv'`
* * The error message should be clear as to where Gauge is looking for the table.
* * Example: `Error: Gauge could not find the table at /env/bar/data2/bar.csv`
* This is a minor point, but there is no need for another property file `data.properties`
* * There are already too many property files...
* * You used `data.properties` to demonstrate the property `gauge_data_dir` in the sample project.
* * This was probably done to highlight the new property, but i just want to have it clarified that a new property file is not needed. 
* * I put the property in `default.properties` and it works, Which is good.
* * Put in your documentation that this property can be put in any of the existing property files including `default.properties` and `java.properties` and gauge will pick it up.

Other than that, things look good. This will solve a lot of our problems.
I will play around some more (if I have time). I use Gauge with `Java`, `Maven` and `Spring`, so I will try to use this build with my own sample project.

**When do you expect to have this feature released?**

## Now, if you can just add the `table` feature at the scenario level, it would be perfect.
This way, we can have different number of rows/different csv files for each scenarios if needed. 
```
## This is a scenario
table: bar.csv

* Show me the <name> and the <email> address.
```
Thanks A lot! Great Work!

sriv:
Thank you for the feedback. To answer your points:

> Gauge should output the path to the table being used during execution on the console.
Yes, I'll add that in. I think this is best printed in `debug` level, since this information is probably needed when you need to troubleshoot. This will also be consistent on how gauge prints out other information like env properties in debug mode.

> When Gauge is unable to find the Table, the error is not clear.
Absolutely, this is a good point. I'll add this information in the error message.

> but there is no need for another property file `data.properties`
True. The properties files are split just to help organize better. You could have them all in a single file and gauge will still pick all the values. As with other properties, you can also set this as an env variable. 

> When do you expect to have this feature released?
Most likely immediately after I have this PR reviewed, though there are a couple of other things that I'd liked to be included.

> if you can just add the table feature at the scenario level, it would be perfect
This feature already exists! Although it is labeled as an experimental feature (mostly because the reporting around this is to be implemented). See https://docs.gauge.org/writing-specifications.html#table-driven-scenario

alpha1592:
@sriv Thanks for taking my feedback. I appreciate it. :)

Yes, the documentation currently states that CSV files or Table notation is not supported on the Scenario level. 
The inline tables are of little use to us since our tables get really large (many rows and columns). We hold our test data and some of assertions in there as well. Plus inline tables are "hardcoded" which makes it difficult to run tests from env to env. 

I would really like to be able to add different types of tests related to the same feature in one Test Specification. 
For example, if testing login page of a web app, you can have the following scenarios..

* Login Successful (table with many rows that represent various roles)
* Invalid login attempts (table with few rows)
* Test Forgot my password link (No table)
* Test Register link (No table)

This is currently not possible if you have one CSV at the Specification level. 
I know that feature is probably a different Feature request, but I wanted to bring attention to it. 

Thanks

sriv:
>  I know that feature is probably a different Feature request, but I wanted to bring attention to it.

Yes, that should be a separate feature request. TBH, I haven't assessed the magnitude of this change. Please do create an issue, so that we do not lose this discussion.

util/fileUtils.go
```diff
@@ -228,7 +228,15 @@ func GetPathToFile(path string) string {
 	if filepath.IsAbs(path) {
 		return path
 	}
-	return filepath.Join(config.ProjectRoot, env.GaugeDataDir(), path)
+
+	gaugeDataDir := env.GaugeDataDir()
+	if gaugeDataDir != "." && filepath.IsAbs((env.GaugeDataDir())) {
+		logger.Warningf(true, "'gauge_data_dir' property should be relative to Project Root. Found absolute path: %s", gaugeDataDir)
```
zabil, line 234:
```suggestion
		logger.Warningf(true, "'gauge_data_dir' property must be relative to Project Root. Found absolute path: %s", gaugeDataDir)
```

zabil (changes requested):

sriv:
removed 'RC' label. This will be clubbed with #2057 in a single release.

zabil (approved):

PR #2081, opened 2021-08-16
Exit with non-zero code if docs plugin has error

### LINKED ISSUE #2080: Documentation plugins should have a non-zero exit code when they return an error
**Describe the bug**
At the moment [documentation plugins exit with a zero exit code even if the plugin explicitly returns an error][1]. This means that in a CI/CD pipeline, for instance, we can't check the exit code and instead have to check the console or log output, which is brittle.

[1]: https://github.com/getgauge/gauge/blob/5cc544c7702df7d3604eaefa6c967eb9a54a2462/plugin/plugin.go#L404-L407

**To Reproduce**
Steps (or project) to reproduce the behavior:
1. Initialise a gauge project
2. Install any gauge documentation plugin
3. Setup the documentation plugin in an invalid state, e.g. missing a mandatory configuration variable
2. Run the gauge docs command
3. See that gauge exits with a zero exit code (instead of a non-zero code, which it should)

```
$ gauge init java # or js, python etc
$ gauge docs <any-documentation-plugin>
```

**Expected behavior**
Gauge should exit with a non-zero exit code when documentation plugins explicitly return an error.

**Versions:**

```
gauge -v
```
```
Gauge version: 1.4.0
Commit Hash: 5cc544c
```

**Additional context**
Proposed amendment would be to use `logger.Fatalf` instead of `logger.Errorf`:

```go
_, err = p.DocumenterClient.GenerateDocs(context.Background(), getSpecDetails(specDirs))
grpcErr := p.killGrpcProcess()
if grpcErr != nil {
    logger.Errorf(false, "Unable to kill plugin %s : %s", p.descriptor.Name, grpcErr.Error())
}
if err != nil {
    logger.Fatalf(true, "Failed to generate docs. %s", err.Error())
}
```

I'm happy to submit a PR if this issue is accepted.

johnboyes:
Prior to this commit running a documentation plugin resulted in a zero
exit code even if the plugin explicitly returned an error. This meant
that in a CI/CD pipeline, for instance, the pipeline couldn't check the
exit code and instead had to check the console or log output, which was
brittle.

Fixes #2080

Signed-off-by: John Boyes <154404+johnboyes@users.noreply.github.com>

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -402,12 +402,12 @@ func GenerateDoc(pluginName string, specDirs []string, startAPIFunc func([]strin
 			logger.Fatalf(true, " %s %s. %s", pd.Name, pd.Version, err.Error())
 		}
 		_, err = p.DocumenterClient.GenerateDocs(context.Background(), getSpecDetails(specDirs))
-		if err != nil {
-			logger.Errorf(true, "Failed to generate docs. %s", err.Error())
+		grpcErr := p.killGrpcProcess()
+		if grpcErr != nil {
+			logger.Errorf(false, "Unable to kill plugin %s : %s", p.descriptor.Name, grpcErr.Error())
 		}
-		err = p.killGrpcProcess()
 		if err != nil {
-			logger.Errorf(false, "Unable to kill plugin %s : %s", p.descriptor.Name, err.Error())
+			logger.Fatalf(true, "Failed to generate docs. %s", err.Error())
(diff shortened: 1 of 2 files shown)


johnboyes:
NB I haven't bumped the Gauge version as thought it better to leave that to the reviewer.

sriv (approved):

sriv (approved):

Issue #2311, opened 2022-12-09
Concepts found in entire project_root


MLaeuchli:
**Describe the bug**
According to the gauge documentation, concepts have to be placed in the directory <project_root>/specs and its subdirectories.
Link: https://docs.gauge.org/writing-specifications.html?os=macos&language=javascript&ide=vscode#concepts
In my test it was not restricted to <project_root>/specs but seems to include the entire <project_root> and subdirectories.

My Project copies specs and concepts and replaces some values before running, this leads to the problem that duplicated concepts are found.
If I missed an option to restrict concept folders, please let me know, that would solve my issue.

**To Reproduce**
Clone or download this repo: https://github.com/MLaeuchli/gauge_concept_search
run
```
npm install
npm run gauge
```

**Logs**

```
[ParseError] <project_root>/specs/Concept.cpt:1 Duplicate concept definition found => 'Search <term> on Wikipedia'
[ParseError] <project_root>/someOtherFolder/Concept.cpt:1 Duplicate concept definition found => 'Search <term> on Wikipedia'
```

**Expected behavior**
The concept in <project_root>/someOtherFolder/Concept.cpt should not be found and therefore not interfere.
Alternatively, a property to restrict where Concepts are allowed could be introduced.

**Versions:**
  
From package.json 
```
"dependencies": {
    "@getgauge/cli": "^1.4.3",
    "taiko": "^1.3.4"
  }
```
From manifest.json
```
{
  "Language": "js",
  "Plugins": ["html-report", "screenshot"]
}
```
 - OS MacOS 13.0.1 (22A400)
 - Runner info Node version: v16.13.1, npm 8.1.2
 - IDE info Intellij IDEA

MLaeuchli:
Hi @haroon-sheikh, thanks for the replay. We found a workaround, but with this change we can solve the problem more elegantly and more in line with how we treat specs.
Sadly I can't test it since I no longer work for that company, but I have forwarded it to them.

PR #2379, opened 2023-07-19
Adds gauge_concepts_dir for defining concepts directory 

haroon-sheikh:
Fixes https://github.com/getgauge/gauge/issues/2311

By default, gauge searches with both `<project root>` and `gauge_specs_dir` for searching and validating the concepts. 

This change is adding a new `gauge_concepts_dir` property so that so we can limit the concepts validation to specific directories. If this isn't set, then current behaviour applies for backward compatibility.

Documentation: https://github.com/getgauge/docs.gauge.org/pull/338

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Install Gauge (windows)
         if: matrix.os == 'windows-latest'
         run: |
-          go run build/make.go --install  --verbose
+          go run build/make.go --install --verbose
           echo "C:\\Program Files\\gauge\\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
       - name: Install Gauge (linux)
@@ -123,7 +123,7 @@ jobs:
       - name: Install Gauge (windows)
         if: matrix.os == 'windows-latest'
         run: |
-          go run build/make.go --install  --verbose
+          go run build/make.go --install --verbose
           echo "C:\\Program Files\\gauge\\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
       - name: Install Gauge (linux)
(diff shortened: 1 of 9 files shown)


haroon-sheikh:
httpUtils tests are currently failing on windows only due to changes in latest golang release 1.20.6. Possibly related to https://github.com/golang/go/issues/61076

chadlwilson:
I guess we could temporarily lock at `1.20.5` so you can move forward?

haroon-sheikh:
@chadlwilson I looked at pinning to 1.20.5 also but it was still failing. After reverting all of my changes and adding smaller bits I've found out its this test thats causing the panic [`6fa7939` (#2379)](https://github.com/getgauge/gauge/pull/2379/commits/6fa7939460f7d6e0a9b4df8fc4a08d30cf982811)

Any ideas what could be going wrong?

cemerick1:
I am just curious the use case of this?

haroon-sheikh:
Hi @cemerick1, as described in the issue. By default, gauge scans for concepts within the full project. This allows us to have a completely separate directory for concepts including outside of the project.

cemerick1:
Thanks @haroon-sheikh for the quick response! Reason I ask is because we would like to limit or completely remove concept steps from appearing in our intellisense/code completion without losing the ability to use them as we are trying to keep it simple.  Does this new feature grant us that ability?

haroon-sheikh:
Yes I think so. If you set the gauge_concepts_dir property to some other directory other than specs/ then it shouldn't include them.

cemerick1:
Good deal! We will give this a whirl on Monday and I will report my findings!

haroon-sheikh:
Actually just thinking about it in detail, it will still show up suggestions for code completions as they're just steps for use in specs. 

It might be worth trying it out anyways. Feel free to raise a separate issue if it doesn't work and we could explore options for you.

cemerick1:
Appears it didn't work.  I have submitted a new issue here  #[2383]

PR #2398, opened 2023-08-22
Fix method comment

peschee:
The method `GetConceptsPaths` does not return `specs` if the concepts dir is not set in the current env.

Related to #2311


zabil:
@peschee thanks for this PR. Can you sign your commits?

Refer https://github.com/getgauge/gauge/pull/2398/checks?check_run_id=16110782102

peschee:
Done

haroon-sheikh (approved):

Issue #2419, opened 2023-10-25
Solution to the typescript issue (hopefully)


bockstaller:
This is more of an investigation report than a bug report, so please excuse me for not using the template.

## Issue Description

The gauge ecosystem (`gauge`, `gauge-vscode`) has its issues [1](https://github.com/getgauge/gauge-vscode/issues/735), [2](https://github.com/getgauge/gauge-vscode/issues/636) and [3](https://github.com/getgauge/gauge-vscode/issues/885) with the inofficial `gauge-ts` plugin. 

The common theme across these issues is these three lines.

```
2021/09/13 15:56:41 jsonrpc2: protocol error: read /dev/stdin: resource temporarily unavailable
2021/09/13 15:56:41 jsonrpc2 handler: sending response 0: jsonrpc2: connection is closed
[Info  - 3:56:41 PM] Connection to server got closed. Server will restart.
```

Which are repeated in `lsp.log` for each plugin reload.
This is only an issue for projects using @BugDiver `gauge-ts` plugin.
In previous cases, this behaviour was fixed by tuning the templates `package.json`, which pinned the typescript/ts-node dependencies to old versions.

## Motivation

The blessed configuration got quite old; maybe it is time to upgrade.
So, I set up a test project with current versions and dug in.

## Investigation

Taking a look at the log lines above and this
![](https://github.com/getgauge/gauge/wiki/images/Gauge-Architecture.png)
architecture drawing points towards an issue between `gauge` and `gauge-vscode`. It is the only part where the systems communicate via `stdio` and use `jsonrpc2`.
This means that for some reason, `gauge` isn't able to connect to its `stdin`. 
But this is only the case for `ts` projects.

<details>
  <summary>Dead End / Tangent 1 -> Upgrading `gauge-vscode`</summary>
  So maybe it is a weird interaction between the project under tests `node_modules` and the `gauge-vscode` plugin.
  So I upgraded the `gauge-vscode` plugin (I'll open a PR), got the tests running, and no luck
</details>

Projects using the `python` plugin run smoothly.
So maybe the plugins are launched differently...

<details>
  <summary>Dead End / Tangent 2 -> Plugin launch method</summary>
	  The `launcher.js` script of the `gauge-ts` plugin is quite complicated in comparison to the straightforward shell script used by `gauge-python`.
[...]

PR #2422, opened 2023-10-26
Upgrade getgauge/common to solve typescript server crash issue

bockstaller:
The getgauge/common upgrade solves the issue
where the gauge server crashes whenever a project is loaded that uses the ts plugin.

The cause is described here https://github.com/getgauge/gauge/issues/2419

This commit closes / permanently fixes issues like
https://github.com/getgauge/gauge-vscode/issues/735 https://github.com/getgauge/gauge-vscode/issues/636 https://github.com/getgauge/gauge-vscode/issues/885

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/apoorvam/goterminal v0.0.0-20180523175556-614d345c47e5
 	github.com/daviddengcn/go-colortext v1.0.0
 	github.com/fsnotify/fsnotify v1.7.0
-	github.com/getgauge/common v0.0.0-20211105093448-e0aee1ccb217
+	github.com/getgauge/common v0.0.0-20231026143554-1599dddc7f01
 	github.com/getgauge/gauge-proto/go/gauge_messages v0.0.0-20210112094005-73ca40cb5ac4
 	github.com/golang/protobuf v1.5.3
 	github.com/magiconair/properties v1.8.7
--- a/go.sum
+++ b/go.sum
@@ -21,13 +21,10 @@ github.com/envoyproxy/go-control-plane v0.9.7/go.mod h1:cwu0lG7PUMfa9snN8LXBig5y
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
 github.com/fsnotify/fsnotify v1.7.0 h1:8JEhPFa5W2WU7YfeZzPNqzMP6Lwt7L2715Ggo0nosvA=
 github.com/fsnotify/fsnotify v1.7.0/go.mod h1:40Bi/Hjc2AVfZrqy+aj+yEI+/bRxZnMJyTJwOpGvigM=
-github.com/getgauge/common v0.0.0-20211105093448-e0aee1ccb217 h1:eaanWSt7UVhb7ibkVLeAyJe3frTi6ixgv/GPfdnEe2U=
-github.com/getgauge/common v0.0.0-20211105093448-e0aee1ccb217/go.mod h1:e3V+gYeNMZt9gGaHqxwnVAwrewx6uauCqT+IE0vZhM8=
 github.com/getgauge/common v0.0.0-20231026143554-1599dddc7f01 h1:YS9Wlu7GKlfQcPzs1tTQ30nsjeAdJeJUqQq+j35YpVo=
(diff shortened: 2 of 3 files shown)


bockstaller:
Sorry @chadlwilson, I know you are still sanity testing ruby but I couldn't resist opening the PR after that debugging journey 😅

chadlwilson:
The plugin functional tests for PRs seem to run with the `master` gauge version. With that in mind, do you need a proper gauge release _before_ you try and get other plugin changes made?

chadlwilson (dismissed):

bockstaller:
Getting these changes released is currently everything that I need to unblock us on our product side.
The optional upgrades to gauge-vscode and gauge-ts are more nice to haves I produced on my way and won't require another gauge release

chadlwilson:
OK, not sure I understand the gauge architecture in sufficient detail to intuitively understand why it's optional on the plugin side of things, so will wait for some input from more expert folks :-)

bockstaller:
I'd like to explain and get corrected if I am wrong.

Gauge uses stdio based jsonrpc2 as the communication channel when used as a language server, as it is the case when used with the vscode extension.
jsonrpc2 expects to have access to stdin whenever it starts up.

The patched prepareCommand function is used in gauge to launch the plugins.
This had the effect that the stdin filedescriptor intended for jsonrpc2 communication got passed on to the plugins gauge launched.

The intention was to allow for debugging with gauge-ruby, but was never used as gauge ruby has its own launcher functionality that passes on os.Stdim

This is only problematic in the case of  node.js as runtime used for gauge-ts. As it is the only runtime that connects to the stdin on plugin launch. 
This happens in a way that prevents jsonrpc2 to use stdin, which leads to the crash of the gauge server.

All other runtimes used for gauge plugins behave in a nicer way and don't lock out jsonrpc2.

Therefore the only change that could be possible would be to configure the node.js runtime to not have this sideeffect, but this isn't possible as far as I am aware off.

haroon-sheikh:
@bockstaller can you bump the gauge version as described in Contributing document please, we can then merge and release with the common change.

haroon-sheikh (approved):

chadlwilson (approved):

chadlwilson:
@haroon-sheikh I've done this on his behalf since I had the PR locally anyway - will shepherd the tests and merge shortly.

chadlwilson:
Released as `1.5.6` - enjoy (?!)

sriv:
> OK, not sure I understand the gauge architecture in sufficient detail to intuitively understand why it's optional on the plugin side of things, so will wait for some input from more expert folks :-)

@chadwilson - not sure I understand the question completely...

chadlwilson:
@sriv Sorry for that. I just meant that I wasn't clear why fixing the issue only required releasing the main gauge (server) with an updated `gauge-common` `prepareCommand` implementation, and why the various plugins didn't also **need** to be updated to fix the underlying issue.

But I infer now that updating the other plugins to use latest `common` is optional insofar as it'd confirm that none of them (accidentally) rely on the [old StdIn redirect logic](https://github.com/getgauge/common/commit/b797a8baab998a6d0a0153162fb0bfa5e81267be) that is now removed, since they all may use `Execute*Command` from `common` for different purposes.

If we don't update them now, we may discover unintentional regressions later?

sriv:
@bockstaller [commented](https://github.com/getgauge/gauge/pull/2422#issuecomment-1781604663) their understanding. I just want to say that this analysis is correct. But also wanted to point out that the use of jsonrpc was something we inherited from the lsp/json-rpc specification. AFAIK the initial lsp ecosystem was jsonrpc over stdio (or maybe we missed the other communication modes). But I do see support for pipes and socket in the [current LSP spec] (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#implementationConsiderations) 

I reckon this is a potential change that can be done at gauge's end.

chadlwilson:
OK, yes, I read his reply, which is why I merged and released :-)

Just to be clear - do you have any further concern with the impact of this change? 

Or particular plugins I can help mitigate risk on by actively updating them to latest gauge-common lib version? Only done `gauge-ruby` of the go-based plugins thus far.

sriv:
It's been a while, but if I believe my memory this specific change was done only for debug support in Ruby and that too for `pry` like cli debugging and not IDE debugging which works differently. 

So, I am going to bet that this won't have an impact on other plugins.

chadlwilson:
Your memory is certainly correct as to the initial motivation so fingers crossed nothing else subsequently relies on those stdin semantics.

https://github.com/getgauge/common/commit/16d1f84d7248590955440a4027cb062c4289a565

PR #2455, opened 2024-01-26
feat: add version parameter for download script

olwe0002:
The current download script is always downloading the [latest](https://github.com/getgauge/gauge/blob/62b9937c03b9d29d2e9f94c0778defce9afc6c4c/build/download.sh#L16) available release.
This MR adds the optional possibility to specify the version you want to install explicitly. The version defaults to `latest` , if its not provided.

```
--version=v1.5.6
```

### DIFF
--- a/build/download.sh
+++ b/build/download.sh
@@ -12,11 +12,17 @@ download() {
     tmp="$(ensure mktemp -d)"
     ensure cd "$tmp"
 
-    # Download latest Gauge zip installer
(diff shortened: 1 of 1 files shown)


sriv (approved):

sriv:
Thanks @olwe0002 . I omitted the PR checks since this is only a change in download.sh, which does not have any tests

PR #2464, opened 2024-02-10
Fix PyPi release

chadlwilson:
PyPi now requires use of API tokens (and 2FA for the user that generates them). https://blog.pypi.org/posts/2024-01-01-2fa-enforced/

Steps needed
- [x] log in and generate an API token
- [x] Add GitHub secret for `PYPI_API_TOKEN` with the token

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -161,15 +161,14 @@ jobs:
 
       - name: Deploy to pypi
         env:
-          PYPI_USER: ${{ secrets.PYPI_USER }}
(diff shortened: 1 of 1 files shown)


chadlwilson:
Validated the approach and twine config w/ API token [via this run](https://github.com/getgauge/gauge-python/actions/runs/7887694484/job/21523522198) --> https://pypi.org/project/getgauge/

So this should be fine for next release.

chadlwilson:
@sriv This can be merged any time, can wait for next release.

PR #2483, opened 2024-02-13
Concept start end notify runner

PiotrNestor:
Update to send concept start and end notification to a runner.
Add isConcept to step data.
Related to the corresponding gauge-proto update

@sriv @jensakejohansson
According to issue: https://github.com/getgauge/gauge-dotnet/issues/203


PiotrNestor:
@sriv 
I cannot set gp.mod to 'go 1.21'. 
It reverts automatically to 'go 1.21.0' which is apparently not acceptable to CodeQL and golangci-lint
Is it because it's 1.21.0 in gauge-proto

PiotrNestor:
@sriv
After gauge-proto 'go 1.21' update

$ go get -u github.com/getgauge/gauge-proto/go/gauge_messages
go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.
go: upgraded go 1.21 => 1.21.0

??

PiotrNestor:
@sriv
Go lint complains that a result should be verified

In notifyBeforeConcept and notifyAfterConcept
   _, err := e.runner.ExecuteMessageWithTimeout(message)
   
Any suggestion what to do with that potential 'err' ?

sriv:
Am yet to go into the details here, but what is the rationale for sending the concept start/end to runners?

PiotrNestor:
@sriv
On the test side it's then possible to produce an informative test trace. 
It provides better understanding of the test flow when a lot of concepts are used.

image

sriv:
I see. But in order for the runner to pass the information to the test suite, you'd need new hooks for Before concept and After Concept with possibly some filtering?

I am not at my computer, but if you have any sample on how your test suite would consume this, perhaps I can offer some thoughts.

One thing to remember is that if gauge starts sending this message, all runners would get it and like we saw in the case of reporting plugins, we will break a lot of other runners since forward compatibility isn't yet there. The way to handle this is that runners have a feature flag capability in the json under 'capabilities'. We will need to check which runner is capable of receiving Concept hooks and send the message only for those which are ready.


Other than that, we may not need to use the ExecuteMessageWithTimeout directly, we can consider ExecuteAndGetStatus or an even more appropriate abstraction. I guess we can borrow the mechanism for steps and use it here

PiotrNestor:
@sriv
This is discussed here:
https://github.com/getgauge/gauge-dotnet/issues/203

The present 'gauge-dotnet' update is here:
https://github.com/system-verification/gauge-dotnet/tree/concept-start-end-notify-runner

Presently, the implementation in the runner is:
- accept the the Concept Start and End notifications
- call the existing Step related hooks
- add 'isConcept' == true if it's a concept step, == false otherwise
- also 'ErrorMessage' and 'StackTrace' are added to the StepInfo

Creating Concept related hooks would result in additional changes without providing better functionality or solution,

jensakejohansson:
To elaborate a bit on @PiotrNestor answer above.

We have quite comprehensive test framework that produces a lots of informative logs about status of the test flows. We also have small "generic" steps that we potentially reuse many, many times within the _same test scenario_ with differnt input data. In the log/trace we log the name of the current executing step using the (BeforeStep hook). Then the following log lines are then related to that step until next step name occurs, fine.

However, since the same step can be executed many times in a scenario and possibility be nested within a concepts it's not as trivial as it seems to easy correlate a log line to a specific point in the spec/flow. If we can log also what concept (potentially nested of course) the step was called from we can easily look at the log and see exactly "where we are" in the scenario.

The screenshot above illustrates how we can now log that a step is part of concept, that is part of an other concept.

We though about a new hook, but decided against the complexity and now just added field that indicates if a step is concept or "real step". Then we catch this information in the BeforeStep hook. It may affect users of course, if the want to perform an action on steps but not concepts they have to filter based on the value of the IsConcept-field.

image

PiotrNestor:
@sriv
We'll proceed with the following updates then:
1. ExecuteAndGetStatus is used to send the Concept stop and end notifications
2. Add ConceptMessages flag to runner info in order to check if the runner supports these. This prevents sending the concept notifications to runners that don't know / want these.
3. Add hooks: BeforeConcept, AfterConcept. Use these for the Concept notifications in gauge-dotnet
4. Also, as a consequence isConcept is removed from gauge-proto StepInfo

go.sum
```diff
@@ -11,10 +11,10 @@ github.com/dmotylev/goproperties v0.0.0-20140630191356-7cbffbaada47 h1:sP2APvSdZ
 github.com/dmotylev/goproperties v0.0.0-20140630191356-7cbffbaada47/go.mod h1:f2V6964+f0p8Asqy8mIK5cKyyVc6MP9PFzGVNRcnYJQ=
 github.com/fsnotify/fsnotify v1.7.0 h1:8JEhPFa5W2WU7YfeZzPNqzMP6Lwt7L2715Ggo0nosvA=
 github.com/fsnotify/fsnotify v1.7.0/go.mod h1:40Bi/Hjc2AVfZrqy+aj+yEI+/bRxZnMJyTJwOpGvigM=
-github.com/getgauge/common v0.0.0-20231211152919-94c93e29f0b9 h1:OwggKdL8AeteB2y0xJzloieq9CMhYG7YmKKMh6D8FLs=
-github.com/getgauge/common v0.0.0-20231211152919-94c93e29f0b9/go.mod h1:p/Q0+qO2bLq08PuxaHrxIgkvKlGGYHyXad33+zKIiXU=
-github.com/getgauge/gauge-proto/go/gauge_messages v0.0.0-20240122132601-bb92eb77d703 h1:ZFK55KGnHAFpwk4YbQsp8r1ybf76MQcEVba5V21qB7w=
-github.com/getgauge/gauge-proto/go/gauge_messages v0.0.0-20240122132601-bb92eb77d703/go.mod h1:grwwUhEUaDZHOj11RDIk5w7tciDv3fYDXeyYoL2BgGc=
+github.com/getgauge/common v0.0.0-20231026143554-1599dddc7f01 h1:YS9Wlu7GKlfQcPzs1tTQ30nsjeAdJeJUqQq+j35YpVo=
+github.com/getgauge/common v0.0.0-20231026143554-1599dddc7f01/go.mod h1:p/Q0+qO2bLq08PuxaHrxIgkvKlGGYHyXad33+zKIiXU=
```
chadlwilson, line 15: Any reason for downgrading `common` here?

chadlwilson:
I haven't followed exactly what we are doing here, but given we broke compatibility with all the reporting plugins aside from `html-report` with `1.5.7` and haven't yet fixed this across all report plugins, we should definitely not break compatibility with _any_ of the language runners or block them from `gauge-proto`s - and should try and validate these.

Even with the feature flags, I would suggest we merge this without an immediate release so that we can then run builds across a few more of the various runners (which use `gauge@master` via [gauge-tests](https://github.com/getgauge/gauge-tests)) to gain confidence that they have not been broken? I believe the FTs run here only validate `gauge-js` and `gauge-java`.

go.sum
PiotrNestor: I'm not editing 'go.sum' directly
'go mod tidy' generated based on 'go.mod'

What to update then?

chadlwilson: I'm not sure what you executed or which revision you started from (possibly a merge/rebase from master problem), but `go mod tidy` won't do that on its own. In effect it's a downgrade (Dec 11 -> Oct 26) which is probably unintentional - no reason to go backwards.

You should probably rebase off master and `go get -t -u ./... && go mod tidy` to get it back to what you'd expect.

PiotrNestor: Bumped back to: github.com/getgauge/common v0.0.0-20231211152919-94c93e29f0b9

execution/scenarioExecutor.go
```diff
@@ -145,12 +145,18 @@ func (e *scenarioExecutor) notifyBeforeConcept(conceptResult *result.ScenarioRes
 	message := &gauge_messages.Message{MessageType: gauge_messages.Message_ConceptExecutionStarting,
 		ConceptExecutionStartingRequest: &gauge_messages.ConceptExecutionStartingRequest{CurrentExecutionInfo: e.currentExecutionInfo, Stream: int32(e.stream)}}
 	e.pluginHandler.NotifyPlugins(message)
+	if (e.runner.Info().ConceptMessages) {
+		_ = e.runner.ExecuteAndGetStatus(message)
```
sriv, line 149: The only feedback I have here (and I apologize for dragging this) is that these ConceptHooks are not sending any result back to the runner. As a result there will be no failure reported, no messages or screenshots will be accumulated.

This is perhaps obvious, but I want to make sure that you are aware of this.

sriv (commented):
The only feedback I have here (and I apologize for dragging this) is that these ConceptHooks are not sending any result back to the runner. As a result there will be no failure reported, no messages or screenshots will be accumulated.

This is perhaps obvious, but I want to make sure that you are aware of this.

This is not a show stopper as these can be added subsequently and need not make it in this PR. Please let me know if you are OK with this behaviour. Other than this this PR LGTM

execution/scenarioExecutor.go
PiotrNestor: We're aware that presently there is no response sent to the gauge runner. An exception in the Concept hooks results only in a (Debug) log output.

It is possible to update gauge-proto NotifyConceptExecutionStarting NotifyConceptExecutionEnding with ExecutionStatusResponse and deliver the proper response.
The question is then how Reporters will handle errors on this Concept step level?

**We suggest to merge this PR. 
We propose to investigate and solve the error management in a subsequent implementation.**

sriv (approved):

chadlwilson:
_(Off-topic FYI)_ The Python and Chocolatey releases worked fine after the release triggered by this, which is good.
https://community.chocolatey.org/packages/gauge
https://pypi.org/project/getgauge-cli/

PR #2492, opened 2024-02-21
Concept hooks return status response

PiotrNestor:
@sriv 
Adding management of concept hooks returning status response.

The related updates are:
- gauge-proto: https://github.com/system-verification/gauge-proto/tree/return-concept-status
- gauge-dotnet: https://github.com/system-verification/gauge-dotnet/tree/return-concept-status

For some reason there is no response / result returned by 'e.runner.ExecuteAndGetStatus' in notifyBeforeConcept, notifyAfterConcept even if the 'gauge-dotnet' has been updated to return a result.

grpcRunner executeMessage (that is called by ExecuteAndGetStatus) returns response == nil

**What might be the reason for that?**
What to check?


PiotrNestor:
@sriv 
OK, I found a problem in 'invokeServiceFor' not returning the correct response :(

sriv:
>  The question now is what to do in 'executeConcept' when there is a Concept start / end response error

I guess we can mark the concept execution result as failed like how its done here:  https://github.com/getgauge/gauge/blob/master/execution/result/conceptResult.go#L53-L66

I am fairly certain that spec/scenario hooks behave similarly, i.e. mark them as failed when the hooks have failures.

PiotrNestor:
@sriv

Simulating a 'BeforeConcept' hook results in the following html-report
The error information is provided but it's not entirely clear that the concept is failing

image

Simulating a 'AfterConcept' hook results in the following html-report
The error information is provided but it's not entirely clear that the concept is failing

image

sriv:
Thanks for testing this out and the feedback. I believe your feedback is around the placement of the error message?

html-report renders concepts as an expandable step with the actual steps listed under it. I guess you would like the errors in the concept hooks should be adjacent to this expandable concept?

This should be possible by changing the template for html-report. The default template lies here https://github.com/getgauge/html-report/blob/master/themes/default/views/partials.tmpl

This uses golang's stdlib templating and the concept container is here: https://github.com/getgauge/html-report/blob/master/themes/default/views/partials.tmpl#L687-L698

If you want to take a stab at this please feel free.

sriv:
Done, merged the corresponding gauge-dotnet and gauge-proto PRs

PiotrNestor:
@sriv 
This PR is ready ...

sriv (approved):

PR #2496, opened 2024-02-29
Added scenario retries info params

PiotrNestor:
@sriv 

Support for scenario retries data in ExecutionInfo.
This provides scenario execution retries information in the hooks context.

This is used to decide if:
- this is the normal scenario execution
- this is a retry of the scenario execution. In this case it's used to possibly reset test data for the scenario.

Some other notes related to '--max-retries-count'
- MaxRetries is '0' if  '--max-retries-count' is not set
- MaxRetries is 1 less than '--max-retries-count' if set
- MaxRetries indicates the actual **retry** count
- CurrentRetry is 0 for the first normal scenario execution

Related updates in:
https://github.com/system-verification/gauge-proto/tree/retries-info-in-context
https://github.com/system-verification/gauge-csharp-lib/tree/retries-info-in-context
https://github.com/system-verification/gauge-dotnet/tree/retries-info-in-context

The updated 'gauge' runner tested with unmodified plugin runners and reporters. These work OK


PiotrNestor:
@sriv
Can we move forward with this feature?

sriv:
yes, looks like a reasonable change proposal.

execution/specExecutor.go
```diff
@@ -299,6 +299,10 @@ func (e *specExecutor) executeScenario(scenario *gauge.Scenario) (*result.Scenar
 			Name:     scenario.Heading.Value,
 			Tags:     getTagValue(scenario.Tags),
 			IsFailed: false,
+			Retries: &gauge_messages.ScenarioRetriesInfo{
```
sriv, line 302: could you please add a test for this? See https://github.com/getgauge/gauge/blob/master/execution/specExecutor_test.go#L617 for a reference

sriv (changes requested):

execution/specExecutor.go
PiotrNestor: 'execution/specExecutor_test.go/TestExecuteScenarioShoulHaveRetriesInfo'

sriv (approved):

PR #2512, opened 2024-03-25
Adds the option to specify env for validate

haroon-sheikh:
Relates to https://github.com/getgauge/gauge/issues/2244

- Adds the option to specify env for validate so that it can use the correct gauge_data_dir for validating the specs.

**Note:**
We cannot filter the validation by tags because to filter the specs, we have to parse them which in return throws the validation errors.

PR #2517, opened 2024-04-02
Release version built with Go 1.22.x

chadlwilson:
Bumps some versions of tools used in the release process to latest LTS.

Note this will trigger release of https://github.com/getgauge/gauge/compare/v1.6.4...master 

Input from @sriv or @haroon-sheikh appreciated!


haroon-sheikh:
Lgtm 👍 let's try to do these version update release every quarter. 

Id also like to include this change as part of 1.4.5 https://github.com/getgauge/gauge/pull/2512

haroon-sheikh:
@chadlwilson just in case you haven’t seen already, the pypi release step is failing.

PR #2589, opened 2024-08-14
Fix cli npm package installs on Linux/MacOS

chadlwilson:
Locks adm-zip package at `0.5.14` since `0.5.15` seemed to introduce the problem at https://github.com/cthackers/adm-zip/issues/530

See https://github.com/getgauge/gauge/pull/2332#issuecomment-2288410472 for discussion


zabil (approved):
Although unrelated, do the lint issues need fixing?

zabil (approved):

PR #2638, opened 2024-10-29
add log and continue when step parsing fails for completion #2632

sriv:


chadlwilson:
Thanks!

I saw many adjacent places where errors are beiing swallowed in this parsing code (in ways that would cause crashes on nil) so I began to assume I was missing something.

sriv:
Yeah, i suppose that is a bad idea. I will give it another go when I find a bit more time, but at the least I'd say we should log it

PR #2692, opened 2025-02-03
Added logic adding/not adding empty lines when formatting specs based on flag/settings.

jensakejohansson:
1. Added a flag that makes it optional if you want the formatter to add empty lines around tags & before table definitions in the spec-files.
2. Implementation in server to handle requests from extension regarding changes in settings.json, where you can define the formatters behavior.
3. settings.go hold the 'state' of settings on the server-side, which can extended to hold any other settings added in the future.

To make it work fully a few lines needs to be in the VSC Extension as well, but I hold that until I know if this will be considered.

Further description in issue:
https://github.com/getgauge/gauge/issues/2693

### DIFF
--- a/api/lang/server.go
+++ b/api/lang/server.go
@@ -18,6 +18,7 @@ import (
 	"github.com/getgauge/gauge/api/infoGatherer"
 	"github.com/getgauge/gauge/execution"
 	"github.com/getgauge/gauge/gauge"
+	"github.com/getgauge/gauge/config"
 	"github.com/sourcegraph/jsonrpc2"
 )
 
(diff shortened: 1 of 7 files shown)


sriv (changes requested):
Looks good, perhaps you could add a test or two? - https://github.com/getgauge/gauge/blob/master/formatter/formatter_test.go

I am ok to merge this even without a test, but just thought I'd mention this in case this was an oversight :)

And I get your point around preferences. Readability is a good enough reason for me!

sriv (approved):

jensakejohansson:
Thanks @sriv! No, I did not forget about tests, just waited if you would disregard the change immediately :)
3 Updates in test-file for formatter:
1. Introduced a setup-method called before each test that makes sure the setting SetSkipEmptyLineInsertions = false before any test starts.
2. Updated existing tests to also include tags, to make sure that is formatted correctly in the "default" behaviour. Now I saw that it actually was a special test already for that, but well... I'll let it be now.
3. Added a new test that formats a scenario with the new flag set to true and makes sure it doesn't add any empty lines to spec.

sriv:
Cool! I've labelled this as a release candidate. If you could just bump the version up I'll merge it right away.

jensakejohansson:
Great, I've made a commit now where I updated to version 1.6.13. 👍

jensakejohansson:
Can you merge it now? Seems like an approval is missing.

sriv (approved):

PR #2751, opened 2025-05-20
Don't compile with CGO for Windows

chadlwilson:
CGO cross-compile was enabled in a8365f4ed7e63424f35f3c92dfe9d518555f6491 10 years ago, but without any comment as to why it was needed and no obvious commits around the same time - perhaps Windows native compilation was much less stable at the time (Golang `1.4` era)?

I'm not sure it is necessary anymore, so probably better to remove? WDYT @sriv?

### DIFF
--- a/build/make.go
+++ b/build/make.go
@@ -36,7 +36,6 @@ const (
 	bin               = "bin"
 	gauge             = "gauge"
 	deploy            = "deploy"
-	CC                = "CC"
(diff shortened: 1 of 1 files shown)


sriv:
Indeed, the CGO support was for legacy reasons, and not something I'd assume is required in 2025.

sriv (approved):

PR #2758, opened 2025-06-08
Aggregate pre+post hook messages and screenshots from parallel tests

### LINKED ISSUE #2747: Missing suite hook messages in JSON report
**Describe the bug**

Gauge version: 1.6.14 (currently the latest version)
Gauge messages sent in Hooks @BeforeSuite and @AfterSuite get lost when tests are executed in parallel. The messages are missing in json-report's result.json then.

**To Reproduce**

I prepared a repository to reproduce the issue.

https://github.com/GerdGriessbach/GaugeMissingSuiteHookMessagesDemo

**Expected behavior**

Gauge messages sent in Hooks @BeforeSuite and @AfterSuite can be found in the json report regardless the tests are executed in parallel or not.

**Quickfix**

Both fixes are probably not clean fixes because I'm not familiar with the Gauge code base and Go at all. But it seems to fix at least my problem with the missing messages. At least with this fixes my tests with parallel test exececutions generates correct reports including the missing Gauge messages.

I identified the code which needs to be fixed. In file execution/parallelExecution.go in method aggregateResults the line

```golang
	r := result.NewSuiteResult(ExecuteTags, e.startTime)
```

accidentally ignores all the values stored in e.suiteResult.

Quickfix 1 (copies all values from e.suiteResult):

```golang
func (e *parallelExecution) aggregateResults(suiteResults []*result.SuiteResult) {
	r := e.suiteResult
	for _, result := range suiteResults {
...
```

Quickfix 2 (copies at least the hook messages):

```golang
func (e *parallelExecution) aggregateResults(suiteResults []*result.SuiteResult) {
	r := result.NewSuiteResult(ExecuteTags, e.startTime)
	r.PreHookMessages = e.suiteResult.PreHookMessages
	r.PostHookMessages = e.suiteResult.PostHookMessages
	for _, result := range suiteResults {
```

chadlwilson:
Fixes #2747 
- should fix for both the "legacy" parallel and the multithreaded grpc case

Not really sure I know what I am doing here or what the consequences might be of this, but this does seem to work in both cases.

Would probably be better to have separate implementations for producing the final suiteResult between the two styles as the `aggregateResults` code path is clearly intended for the original legacy design.

### DIFF
--- a/execution/parallelExecution.go
+++ b/execution/parallelExecution.go
@@ -295,9 +295,19 @@ func (e *parallelExecution) finish() {
 
 func (e *parallelExecution) aggregateResults(suiteResults []*result.SuiteResult) {
 	r := result.NewSuiteResult(ExecuteTags, e.startTime)
+	if e.suiteResult != nil {
+		r.PreHookMessages = e.suiteResult.PreHookMessages
+		r.PreHookScreenshotFiles = e.suiteResult.PreHookScreenshotFiles
+		r.PostHookMessages = e.suiteResult.PostHookMessages
+		r.PostHookScreenshotFiles = e.suiteResult.PostHookScreenshotFiles
+	}
 	for _, suiteResult := range suiteResults {
 		r.SpecsFailedCount += suiteResult.SpecsFailedCount
 		r.SpecResults = append(r.SpecResults, suiteResult.SpecResults...)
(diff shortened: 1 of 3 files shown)


GerdGriessbach:
This fix looks good to me. Thanks.

sriv (dismissed):

PR #2764, opened 2025-07-15
Use timestamps in RFC3339 format

sschulz92:
With this change I can distinguish between executions more precisely. The current format "Jul 15, 2025 at 9:23am" does not allow me to handle really fast APIs in different workspaces. If the simple removal is not an option, it might be possible to configure the format like we have the other config params too.

JSON report:

image

XML report:

image

HTML report:

image

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -14,7 +14,7 @@ import (
 )
 
 // CurrentGaugeVersion represents the current version of Gauge
-var CurrentGaugeVersion = &Version{1, 6, 19}
+var CurrentGaugeVersion = &Version{1, 6, 20}
 
 // BuildMetadata represents build information of current release (e.g, nightly build information)
 var BuildMetadata = ""


chadlwilson:
I'm not sure if anything relies on it, but changing the format of an existing field like this is really a breaking change IMHO (when changing from a friendly readable date to a more technical format, I mean, not all changes).

Existing consumers (whether plugins or end usages of the plugins) poasibly have a reasonable expectation of the format staying a readable format as it has been.

Would it be wiser to introduce a new field?

sriv:
I think that is reasonable, but I am also wondering if the property needs to be universal? I cannot recall offhand if there are other places where the time format needs to be honoured.

chadlwilson:
I think the problem with making it configurable is that it cannot then be used by plugins as a contract with which to parse the timestamp and display it as appropriate for them (e.g display a time in the user's timezone independent of Gauge timezone), which is one of the advantages of moving this to a proper ISO format with TZ and using such time formats as a reliable means of integration.

sschulz92:
As far as I understood the whole flow, the timestamp is formatted in gauge itself into a string and passed around to whatever plugins you are using. The format you might configure in the project, will be used everywhere out-of-the-box. 

If I add another field, we are talking about the proto message, right? Afterwards I am able to use the field everywhere including gauge-core. To me it's a little bit strange to add another field with the - more or less - exact same information except the accuracy.

sriv:
raised a PR for the proto changes here: https://github.com/getgauge/gauge-proto/pull/111

sschulz92:
Is there something missing in gauge-proto? The generation of the go-files?

sriv:
had to update gauge-proto references in this repo, all good now

PR #2792, opened 2025-10-28
implement lazy initialization strategy for scenarios

sriv:
- **make scenarios initialize lazily for table driven scenarios**
- **add benchmark tests for lazy scenario initialization**


sriv:
@jensakejohansson - thanks for checking this out. I had missed one path with nested tables. I have pushed another commit - think this should address the issue. If you can take another look, that'd be super helpful!

jensakejohansson:
@sriv I don't know if I'm confused here, but I don't see any new commits?

Issue #2790, opened 2025-10-24
Finalize the implemention of table driven scenarios


jensakejohansson:
The support for table driven scenario has been an experimental feature of Gauge since 1.0.3.
https://docs.gauge.org/writing-specifications?os=linux&language=csharp&ide=vscode#table-driven-scenario

It would be good if we could get a grip on this and make a more complete implementation. The goal would be to remove the flag/env: allow_scenario_datatable and make "true" the default. 

@sriv Below is my understanding/thoughts of the current state of things. I think the implementation got pretty far, works better than I expected, but still enough things to do to give one nightmares :)

My thoughts so far:

1. Should it be really be allowed to have table driven scenarios within a table driven spec? 
At the moment it's possible and there are even reported issue: https://github.com/getgauge/gauge/issues/2317
Even though it might be technically possible I would argue no. I believe the purpose of Gauge is to write simple and understandable tests/scenarios. Having tables defined at different levels and possibly nested loops of execution seems to defeat this purpose? Not to mentions challenges in the actual Gauge and report implementations. My suggestion would be to let the parser/validator notify that "nested" tables - if found - are not accepted. Either you have a table on spec-level or scenario-level, you cannot combine in the same .spec-file.

2. Plugin support: 
The docs says IDE plugins does not support this feature. I did not see any issues in VSC - which I primarily use. I tried the plugin for IntelliJ. There it highlights the table as incorrect / unexpected, but execution worked for my simple test scenario. I thought the information came from the language server, so the reason for the difference is unclear to me. That said I have never looked at the implementation for the IntelliJ plugin.

<img width="745" height="290" alt="Image" src="https://github.com/user-attachments/assets/1227c2fa-0718-48ba-b304-ffe0f404ec95" />

Conclusion: IntelliJ plugin needs to fixed.

3. CSV-files cannot be used for table data in table driven scenarios as it can with specs.

<img width="580" height="307" alt="Image" src="https://github.com/user-attachments/assets/b739b40b-e310-49df-8e66-27e1e7031877" />

When trying this I realized I missed the "table:" keyword when implementing the syntax-highlighter in the VSC extension. It's highlighted as a comment. I will correct that.
Conclusion: Need to add support for CSV-files. 

4. Reporting is inconsistent and should also benefit from some improvements:
[...]

PR #2798, opened 2025-11-12
More complete information to reporters when scenarios are table-driven.

jensakejohansson:
The information in the results sent to the reporters (xml/html-report) were incomplete in case of table driven scenarios. This change-set addresses this, making it clear if a scenario was spec-table-driven, scenario-table-driven or possibly both - also including the actual scenario table with a reference to the correct row index.

Part of: https://github.com/getgauge/gauge/issues/2790


jensakejohansson:
@sriv 👍just remember, there are related updates in:
https://github.com/getgauge/xml-report/pull/116

And this, which bigger and potentially a bit more risky:
https://github.com/getgauge/html-report/pull/366

Issue #2855, opened 2026-06-04
Resolving launcher.cmd for dotnet on Windows fails in certain situations


jensakejohansson:
Launching Gauge (dotnet) fails in two cases for me (on Windows):

1. Can no longer launch gauge dotnet runner in VSC via the Gauge extension - this worked before (until today).
2. Launch fails when I ask Claude Code CLI to run gauge, this has never worked for me.

Error message:
```
Running tool: gauge.exe run --hide-suggestion --simple-console c:\Users\JensJ\Projects\temp\temp9\specs\example.spec:19
'launcher.cmd' is not recognized as an internal or external command,
operable program or batch file.
Error occurred while waiting for runner process to finish.
Error : exit status 1
Error ----------------------------------

[Gauge]
Failed to start gauge API: Error occurred while waiting for runner process to finish.
Error : exit status 1

Get Support ----------------------------
	Docs:          https://docs.gauge.org
	Bugs:          https://github.com/getgauge/gauge/issues
	Chat:          https://github.com/getgauge/gauge/discussions

Your Environment Information -----------
	windows, 1.6.30
	dotnet (0.7.7), html-report (4.4.0), java (1.0.1), js (5.0.1), reportportal (1.18.0), screenshot (0.3.2), ts (0.3.6), xml-report (0.5.4)
Error: Tests failed.
```

From the logs:
```
04-06-2026 13:06:24.344 [dotnet] [ERROR] 'launcher.cmd' is not recognized as an internal or external command,
04-06-2026 13:06:24.345 [dotnet] [ERROR] operable program or batch file.
```

<img width="1109" height="1272" alt="Image" src="https://github.com/user-attachments/assets/82ecfe85-5abc-4c08-a840-ebeb8b45b20b" />

<img width="653" height="262" alt="Image" src="https://github.com/user-attachments/assets/245083cc-7f0f-43c8-838c-55a84aa9b2d2" />

Regarding VSC, something obviously 'moved', it is unclear to me what it is. Running gauge from cmd-terminal/git bash works fine.

Since I've seen the issue before inside a Claude Code CLI when I've asked Claude to run gauge, I suspected it was some sort of shell/environment setup that is expected but does not work as intended in all cases.

[...]

PR #2856, opened 2026-06-04
Start lang runner using abs path to mitigate Windows shell issues

jensakejohansson:
PR suggestion to mitigate issues with launching language runners on Windows where .cmd/.bat files are the entry point:
https://github.com/getgauge/gauge/issues/2855

A directory for the command was set in current implementation, but obviously not always honored on Windows. 

Therefore the suggestion is to point to the runners entry point using an absolute path instead. This of course requires that the language runner is designed so that works (i.e. entry point does not refer to some command expected to be in path). For the official runners, as far as I can see, this is not an issue.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -96,6 +96,9 @@ func runRunnerCommand(manifest *manifest.Manifest, port string, debug bool, writ
 		return nil, nil, fmt.Errorf("Compatibility error. %s", compatibilityErr.Error())
 	}
 	command := getOsSpecificCommand(&r)
+	if len(command) > 0 && !filepath.IsAbs(command[0]) {
+		command[0] = filepath.Join(runnerDir, command[0])
+	}
 	env := getCleanEnv(port, os.Environ(), debug, getPluginPaths())
 	cmd, err := common.ExecuteCommandWithEnv(command, runnerDir, writer.Stdout, writer.Stderr, env)
 	return cmd, &r, err
--- a/version/version.go
+++ b/version/version.go
@@ -14,7 +14,7 @@ import (
 )
 
 // CurrentGaugeVersion represents the current version of Gauge
-var CurrentGaugeVersion = &Version{1, 6, 31}
+var CurrentGaugeVersion = &Version{1, 6, 32}
 
 // BuildMetadata represents build information of current release (e.g, nightly build information)
 var BuildMetadata = ""


chadlwilson (commented):
I don't think this is appropriate without understanding the root cause.

To my knowledge there are semantic differences to running things inside a given working directory vs absolute on some platforms, and this seems as likely to get us into a mess as the other issues.

jensakejohansson:
Agree knowing the root cause is preferred, but I have no idea how to proceed so I'll have to leave it as is. Maybe it helps someone.

On the note of working directory, it is still set as before and understood once the .bat/executable actually runs, otherwise that .cmd script would fail since it has relative references.

```
dotnet bin\net10.0\Gauge.Dotnet.dll %*
if %errorlevel% neq 0 exit /b %errorlevel%
goto :eof

:net8
    dotnet bin\net8.0\Gauge.Dotnet.dll %*
    if %errorlevel% neq 0 exit /b %errorlevel%
    goto :eof

:net9
    dotnet bin\net9.0\Gauge.Dotnet.dll %*
    if %errorlevel% neq 0 exit /b %errorlevel%
    goto :eof
```

jensakejohansson:
Got it. Added some logging. Running in CMD works:

```
04-06-2026 20:59:07.852 [Gauge] [DEBUG] NoDefaultCurrentDirectoryInExePath=
04-06-2026 20:59:07.852 [Gauge] [DEBUG] Runner command: [start.bat --start], runnerDir: C:\Users\JensJ\AppData\Roaming\gauge\plugins\python\0.5.0
04-06-2026 20:59:08.572 [Gauge] [DEBUG] Checking updates...
```

When run is triggered by VSC it fails:

```
04-06-2026 20:52:51.985 [Gauge] [DEBUG] NoDefaultCurrentDirectoryInExePath=1
04-06-2026 20:52:51.985 [Gauge] [DEBUG] Runner command: [start.bat --start], runnerDir: C:\Users\JensJ\AppData\Roaming\gauge\plugins\python\0.5.0
04-06-2026 20:52:52.017 [python] [ERROR] 'start.bat' is not recognized as an internal or external command,

```
NoDefaultCurrentDirectoryInExePath is set and it breaks Gauge approach of finding the runner.

Why? I had no idea, so I threw Claude at searching for it and the irony is deep here:

It's Claude Code's extension - according to Claude:

> process.env.NoDefaultCurrentDirectoryInExePath="1"                                                                                                                                                                                                                                                                                            
>    
> It appears twice in anthropic.claude-code-2.1.162-win32-x64/extension.js:                                                                                                                                                                                                                                                                     
> 1. At module initialization (top-level)                                                                                                                                                                                                                                                                                                     
> 2. In the activate function (c34)                                                                                                                                                                                                                                                                                                           
> 
> Claude Code sets NoDefaultCurrentDirectoryInExePath=1 when its VS Code extension activates. This propagates to the entire VS Code process, and all child processes (including gauge.exe) inherit it.
> 
> So the timeline is:
> 1. Installed/updated the Claude Code VS Code extension
> 2. It sets NoDefaultCurrentDirectoryInExePath=1 at activation
> 3. The Gauge extension spawns gauge.exe, which inherits this variable
> 4. cmd.exe /c "start.bat" inside gauge fails because cmd.exe no longer searches CWD

I have verified this.

Since any Node.js-based tool may set this variable for security reasons I don't believe the current approach is viable? We assume the regression is not trustworthy?

chadlwilson:
> Since any Node.js-based tool may set this variable for security reasons I don't believe the current approach is viable? 

If Claude Code's extension is breaking other extensions by mutating global shared environment, that's a design problem in itself, and doesnt seem to be Gauge's problem to me. They could alter the process's `env.PATH` and remove everything including gauge.exe - would that also be Gauge's problem to solve?

`NoDefaultCurrentDirectoryInExePath` is a Windows env var that `cmd` looks at, so it's not specific to Node. Any user could presumably set it in their machine env independent of VSCode and break things.

I don't quite get why we should change the whole launch semantics across OSes, given there have been zero reports of anyone using this ancient env var in Gauge's existence in a way that breaks Gauge, and likely this is happening now because LLM-based tools are rife with possible security problems.

If that's the problem, changing the Windows run commands to `.\launcher.cmd` should work equally well.

> We assume the regression is not trustworthy?

Not sure what this means?

jensakejohansson:
Changing run command to `.\launcher.cmd` did not worked when I tried, but will not investigate. It would also require updates across many runners.

Anyhow, I've understood the issue and can mitigate it by running my own builds (which I need to do anyways due to other issues), or adding the runner dir to path.

Closing this PR and adding a note in the issue so other can find it if they want to use Claude extension or Claude Code (CLI).

chadlwilson:
> Changing run command to `.\launcher.cmd` did not worked when I tried, but will not investigate. It would also require updates across many runners.

That's surprising, I'll have a play on my VM later. I'd rather make mass changes to config across repos to do things the right way (and limit surface area to Windows) than add centralised workarounds.

jensakejohansson:
> > Changing run command to `.\launcher.cmd` did not worked when I tried, but will not investigate. It would also require updates across many runners.
> 
> That's surprising, I'll have a play on my VM later. I'd rather make mass changes to config across repos to do things the right way (and limit surface area to Windows) than add centralised workarounds.

My bad, it took it literally, the backslash must be escaped to be treated correctly, so `".\\launcher.cmd"` seems to work.

chadlwilson:
I think this is one of those things that is not particularly well-understood on Windows, and people have just got used to writing invocations without `.\` and relying on the `PATH` semantics (which they dont on POSIX OSes). It'd be better practice to be using this everywhere, I imagine.

PR #2857, opened 2026-06-10
Add option to configure the maximum of log files to retain

sschulz92:
Currently, the maximum number of gauge.log backup files is hardcoded to 3 in logger/logger.go. There is no way to change this without modifying and recompiling the source. In my code-base I need to have more log files as the project is huge and sometimes it is easier to see the point looking into the log files (and their timestamps).

This PR adds a new `log_max_backups` key to `gauge.properties`, allowing users to configure the number of backup log files to retain. The default value remains 3, so existing behaviour is unchanged.

Users can change the value via the existing gauge config CLI:

```bash
gauge config log_max_backups_count 10
```

Or: they rely on the well-known-default value of 3 without the need to change anything.

### DIFF
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -25,6 +25,7 @@ const (
 	ideRequestTimeout       = "ide_request_timeout"
 	checkUpdates            = "check_updates"
 	allowInsecureDownload   = "allow_insecure_download"
+	logMaxBackups           = "log_max_backups"
 
 	defaultRunnerConnectionTimeout = time.Second * 25
 	defaultPluginConnectionTimeout = time.Second * 10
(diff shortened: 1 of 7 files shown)


sschulz92:
@chadlwilson I just added a brief overview of this PR, mind having a look onto the changes? 😄

config/configuration.go
```diff
@@ -25,13 +25,15 @@ const (
 	ideRequestTimeout       = "ide_request_timeout"
 	checkUpdates            = "check_updates"
 	allowInsecureDownload   = "allow_insecure_download"
+	logMaxBackups           = "log_max_backups"
```
zabil: Can the naming of the parameter be consistent across i.e `logMaxBackupsCount` `log_max_backups` 
or 

`LogMaxBackups()`

sschulz92: Do you prefer the one with or without count at the end? I tend to choose the one with count.

zabil: With the count reads better

zabil (approved):

Issue #1420, opened 2018-02-09
Gauge table in concept file doesn't accept |-------------| after the header.


osandadeshan:
### Expected behavior
Gauge table in concept file should accept |-------------| after the header

### Actual behavior
Gauge table in concept file doesn't accept |-------------| after the header

### Steps to reproduce
1. Create a concept file with a gauge table like below
* And the JSON Path Assertions for the response should be equal to the following
```
     |JSON Path                           |Expected Result                                                                   |
     |------------------------------|----------------------------------------------------------------|
     |$.title                                   |Expert API Automation using MaxSoft ATA Framework      |
     |$.userId                               |osan                                                                                      |
     |$.archived                            |true                                                                                       |
     |$.book.bookTitle                 |Expert API Deck3                                                                   | 
```
2. Create a spec and provide the concept name as a step name
3. Run that spec.
4. You will get an error that |-------------| has taken as a data record.

### Gauge version
```
Gauge version: 0.9.7
Commit Hash: bc9784c

Plugins
-------
csharp (0.10.1)
flash (0.0.1)
html-report (4.0.2)
java (0.6.6)
ruby (0.4.2)
screenshot (0.0.1)
spectacle (0.1.2)
xml-report (0.2.0)
```

singhverse:
I'm having this issue too in **Gauge 1.0.6**.
The workaround is not working for me though. 

@negiDharmendra 
Whats the status on fixing this? Is there a different workaround i can use for now?

Romalouz:
As a workaround, I'm doing this:
```C#
            foreach (var row in table.GetTableRows())
            {
                if (IsInvalidTableRow(row, table.GetColumnNames()[0]))
                    continue;
                // row processing
            }


        private static bool IsInvalidTableRow(TableRow row, string testColumn)
        {
            return row.GetCell(testColumn).Contains("---");//check if gauge is sending the separator row...
        }

```

jensakejohansson:
Yes, this is along time bug. Reason is that conceptParser does not clear states (tableScope / talbleSeparatorScope) when starting a new step. A blank line after the table "accidentally" clears the states - it's considered to be comment as I recall (anything not matching header/step/table) and then all but conceptScope is cleared. Back-to-back steps do not clear the table states.

If it helps anyone, my change to mitigate this was simply to clear the scopes (retain only the conceptScope) in the parser when a new step is started. A one-line-fix in conceptParser.go
```

} else if parser.isStep(token) {
[...]

PR #2860, opened 2026-06-30
Fix for the parse error with back-to-back table tests in concept

jensakejohansson:
# Problem

When a concept definition contains two consecutive steps that each have an inline table (with no blank line between them), the second table's separator row (|--|--|) is incorrectly treated as a data row and appended to the first step's table.

This is been around for a long time (always?): https://github.com/getgauge/gauge/issues/1420

For example:

```
  # my concept
  * step with table
   |id|name|
   |--|----|
   |1 |one |
  * step with table
   |id|name|
   |--|----|       <-- treated as data row, not a separator
   |2 |two |
```

# Why?

The ConceptParser uses a bitmask-based state machine to track the current parsing context. When a table data row is encountered, the parser checks:

```
  if areUnderlined(token.Args) && !isInState(parser.currentState, tableSeparatorScope) {
      addStates(&parser.currentState, tableSeparatorScope)
  } else if isInState(parser.currentState, stepScope) {
      parser.processTableDataRow(token, &parser.currentConcept.Lookup, fileName)
  }
```

The first table's separator sets tableSeparatorScope in the state. When a new step token was encountered, the parser only called addStates(&parser.currentState, stepScope) without clearing the previous state. This meant tableSeparatorScope persisted from the first step into the second.

When the second table's separator `|--|--|` arrives, areUnderlined returns true but `!isInState(parser.currentState, tableSeparatorScope)` is false because it's still set from the first table. So the first branch is skipped, and the else if branch treats it as a regular data row, calling `processTableDataRow` which appends `--|--` as data to the current step's table.

# Fix

Added `retainStates(&parser.currentState, conceptScope)` before `addStates(&parser.currentState, stepScope)` when processing a new step token. retainStates performs a bitmask AND, keeping only the specified scopes and clearing everything else, including the stale tableSeparatorScope (and tableScope) left over from the previous step's inline table. This allows the second table's separator to be correctly recognized.

This is consistent with how state transitions are already handled elsewhere in the parser for example, the comment token handler at line 104 uses the same retainStates(&parser.currentState, conceptScope) pattern.

# Test

* Added `TestParsingConceptWithConsecutiveInlineTablesWithoutBlankLine` which parses a concept with two steps each having an inline table (no blank line separator) and verifies:
  - Parse succeeds with no errors
  - The concept contains exactly 2 steps
  - Each step has its own initialized inline table with the correct columns, row count, cell values, and cell types
* Verified the test fails before the fix and passes after

### DIFF
--- a/parser/conceptParser.go
+++ b/parser/conceptParser.go
@@ -82,6 +82,7 @@ func (parser *ConceptParser) createConcepts(tokens []*Token, fileName string) ([
 			if errs := parser.processConceptStep(token, fileName); len(errs) > 0 {
 				finalResult.ParseErrors = append(finalResult.ParseErrors, errs...)
 			}
+			retainStates(&parser.currentState, conceptScope)
 			addStates(&parser.currentState, stepScope)
 		} else if parser.isTableHeader(token) {
 			if !isInState(parser.currentState, stepScope) {
--- a/parser/conceptParser_test.go
+++ b/parser/conceptParser_test.go
@@ -463,6 +463,56 @@ func (s *MySuite) TestParsingConceptStepWithInlineTable(c *C) {
 	c.Assert(nameCells[1].CellType, Equals, gauge.Static)
 }
 
+func (s *MySuite) TestParsingConceptWithConsecutiveInlineTablesWithoutBlankLine(c *C) {
+	parser := new(ConceptParser)
+	concepts, parseRes := parser.Parse("# my concept\n"+
+		"* step with table\n"+
+		" |id|name|\n"+
+		" |--|--|\n"+
+		" |1|one|\n"+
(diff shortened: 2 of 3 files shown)


jensakejohansson:
I suppose the failure in the regression is due to the known flakiness of the regression we've seen earlier and not due to the actual code change?

```
Failed Step: Console should contain "Successfully generated html-report to =>"
        Specification: specs/advanced_topics/parallel_execution.spec:76
        Error Message: java.lang.AssertionError: 
        Expecting actual:
          "Executing in 2 parallel streams.
        Incorrect value for enable_multithreading in property file. Cannot convert  to boolean.
        Using default value false for property enable_multithreading.
```

chadlwilson (approved):

PR #2878, opened 2026-08-06
Add option to decide on execution strategy when using data tables

sschulz92:
This PR is about to resolve the topic described in https://github.com/getgauge/html-report/issues/388 

You can decide which strategy to use:
1. filter (default): keep everything as it is, scenarios without accessing the data table, are executed once
2. matrix: every scenario of the spec will be executed regardless if the data table is accessed or not

You can configure Gauge with:

```
gauge config data_table_mode <OPTION>
```

I tested it locally, both options, and it is like expected: 4 scenarios by default, 6 scenarios using the new option "matrix". The numbers are explained in the mentioned issue.

### DIFF
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -26,6 +26,7 @@ const (
 	checkUpdates            = "check_updates"
 	allowInsecureDownload   = "allow_insecure_download"
 	logMaxBackupsCount      = "log_max_backups_count"
+	dataTableMode           = "data_table_mode"
 
 	defaultRunnerConnectionTimeout = time.Second * 25
 	defaultPluginConnectionTimeout = time.Second * 10
@@ -34,6 +35,7 @@ const (
 	defaultRunnerRequestTimeout    = time.Second * 30
 	defaultIdeRequestTimeout       = time.Second * 30
 	defaultLogMaxBackupsCount      = 3
+	defaultDataTableMode           = "filter"
 	LayoutForTimeStamp             = "Jan 2, 2006 at 3:04pm"
 )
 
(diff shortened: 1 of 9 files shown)


chadlwilson:
What's the use case? When is this ever desirable or correct? Even if there is a use case I cannot see, I'm not comfortable adding this, as it adds yet another combination to reason about and test.

I would even support failing validation of any scenario that doesn't refer to a spec table that exists (albeit a breaking change).

It seems to go against the decision in https://github.com/getgauge/gauge/issues/219 and this will just increase the number of places we might get things wrong, and the existing alleged bugs when running individual scenarios alongside a spec table would get more confusing to address.

sschulz92:
> What's the use case? When is this ever desirable or correct? Even if there is a use case I cannot see, I'm not comfortable adding this, as it adds yet another combination to reason about and test.
> 
> I would even support failing validation of any scenario that doesn't refer to a spec table that exists (albeit a breaking change).
> 
> It seems to go against the decision in #219 and this will just increase the number of places we might get things wrong, and the existing alleged bugs when running individual scenarios alongside a spec table would get more confusing to address.

It was quite surprising for me to see that test scenarios were not executed, even though they were part of a spec containing a spec table. It was not at all clear to me that Gauge behaves this way - and the behavior is also misrepresented (for instance, in the HTML report). In the example, it is easy to follow because only a few scenarios are implemented. In my own context, however, the situation is different; there are many scenarios across many different specs. I believe I am not alone in this opinion.

Fixing the HTML report is something I will investigate, not sure if I can help though.

chadlwilson:
But why execute a scenario n times when each execution will be identical due to not referring to any table values? What am I missing here?

This seems it would be even more misleading because it looks like each scenario has indeed been varied by the spec table values, when in fact they have not? Instead it's just wasteful?

I really don't see how this is helpful. Perhaps there needs to be a real example to follow?

The fact that the HTML report is currently wrong (and likely has been for years) is irrelevant - there are a number of bugs across across gauge and plugins relating to give both types of tables, probably magnified by their n*m matrixed combination.

chadlwilson:
I must be missing something because it doesn't seem accurate to characterise them as "test scenarios were not executed" - they are executed - just in non data table (normal) mode, correct?

sschulz92:
> But why execute a scenario n times when each execution will be identical due to not referring to any table values? What am I missing here?

I can see your point - from the execution perspective it does make sense to run scenarios once, if they are not using the data table. If I look from the user point of view, reading a spec file, I would expect the matrix nevertheless. Maybe this is exactly what he/she wants. Of course there would be solutions to "overcome" this, by using the data table.

All in all: I was not aware of this scenario-filtering functionality and was tricked by the HTML report. If we enhance the HTML report to show the truth like the JSON report, I am fine!

> I must be missing something because it doesn't seem accurate to characterise them as "test scenarios were not executed" - they are executed - just in non data table (normal) mode, correct?

Those scenarios are executed first, then those accessing the data table. None of them is marked as "not executed". Execution wise, everything is OK.

PR #2884, opened 2026-08-21
Authenticate GitHub plugin downloads with GITHUB_TOKEN

### LINKED ISSUE #2883: gauge plugin installs should respect env GITHUB_TOKEN or standard authentication mechanisms
**Is your feature request related to a problem? Please describe.**

Currently it is quite common for plugin installs to fail, seemingly because non-authenticated Github traffic is heavily rate limited. It's possible this issue is specific to GitHub Actions builds, and the code already supports it, however there should be a standard way to authenticate the traffic where a plugin is hosted on Github.

**Describe the solution you'd like**

- verify whether this is the root cause of the issues on GitHub actions
- Respect the GITHUB_TOKEN in env by default; doing `Bearer` auth with Github.
- Ensure it is not/cannot be logged out anywhere
- Document it accordingly
- Possibly provide an alternative (secure!) mechanism for providing a token to Gauge.
  - Not raw off the file system

**Additional context**

See https://github.com/getgauge/gauge/actions/runs/32472415045/attempts/1 for an example.  It was a gateway timeout rather than explicit rate limit, however I think that is what happens for unauthed traffic now when load-shedding...

```

Run gauge install java
  gauge install java
  shell: /usr/bin/bash -e {0}
  env:
    BENCHMARK_OUT_DIR: ../out
    GOTOOLCHAIN: local
    JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/21.0.12-8/x64
    JAVA_HOME_21_X64: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/21.0.12-8/x64
    MAVEN_ARGS: -ntp
2026/08/21 10:37:41 Failed to get configuration from Gauge properties file. Error: open /home/runner/.gauge/config/gauge.properties: no such file or directory
Failed to install plugin 'java'.
Reason: Failed to download the plugin. Error downloading file: https://github.com/getgauge/gauge-java/releases/download/v1.0.3/gauge-java-1.0.3-linux.x86_64.zip.
504 Gateway Timeout
```

dhananjaypesu:
Closes #2883.

## Problem

`gauge install <plugin>` fetches releases from GitHub unauthenticated, and that traffic is rate limited hard enough to fail on shared CI runners. As the issue notes, it usually surfaces as a `504 Gateway Timeout` while GitHub sheds load rather than an explicit 403 — which is exactly why the failure does not look like a rate limit.

## Fix

`util.Download` builds its request explicitly and adds `Authorization: Bearer <token>` when a token is configured **and** the URL is a GitHub host.

**Two variables**, in precedence order:

| Variable | Notes |
|---|---|
| `GAUGE_GITHUB_TOKEN` | Checked first — the "alternative mechanism" the issue asks about. Lets a CI job give Gauge a token without widening what every other tool in the job sees. |
| `GITHUB_TOKEN` | Used when the above is unset. On GitHub Actions this is the automatic per-job token, so the failing workflow in the issue is fixed by adding one `env:` line. |

A blank or whitespace-only value counts as unset, so `GITHUB_TOKEN: ""` does not produce a `Bearer ` header.

**The host list is deliberately narrow** — `github.com`, `www.github.com`, `api.github.com`, `codeload.github.com`. This is the part worth reviewing:

- Release assets **redirect** to `objects.githubusercontent.com`, which rejects a request carrying an `Authorization` header. `net/http` already drops the header on that cross-host hop, and the allowlist means we never set it for that host in the first place either — belt and braces, because getting this wrong turns a rate-limit fix into a hard 400 on every install.
- Matching is on the parsed **hostname**, not a substring, so `evilgithub.com` and `github.com.attacker.test` get nothing. There are tests for both.

**"Ensure it is not/cannot be logged out anywhere"**: the token is never part of a URL, so the existing `logger.Debugf(true, "Downloading %s", url)` and every error message carry the URL only, and nothing in this path prints request headers. There is a test asserting a failed download's error does not contain the token.

## Tests

Seven new cases in `util/httpUtils_test.go`:

- the token is returned for all four GitHub hosts, and host matching is case-insensitive;
- it is **not** returned for `objects.githubusercontent.com`, an unrelated host, the two lookalike hosts, localhost, or an unparseable URL;
- absent when neither variable is set; `GAUGE_GITHUB_TOKEN` beats `GITHUB_TOKEN`; a whitespace-only value falls through;
- an end-to-end `Download` against an `httptest` server asserts **no** `Authorization` header reaches a non-GitHub host — the regression that would break every non-GitHub download;
- a failing download's error does not contain the token.

A small `setEnv` helper returns a restore function, since gocheck's `*C` has neither `Setenv` nor `Cleanup`.

## Verification

```
go build ./...                    # ok
go vet ./util/                    # clean
gofmt -l util/                    # clean
go test ./... -count=1            # exit 0
```

## Docs

README gains an **Authenticating plugin downloads** section: both variables, the read-only scope that suffices, a GitHub Actions snippet, and where the token is and is not sent. If the user-facing home for this is [docs.gauge.org](https://github.com/getgauge/docs.gauge.org) instead, point me at the page and I will send that PR too.


chadlwilson:
No thanks. Design first, Pr later. See https://github.com/getgauge/gauge/issues/2883#issuecomment-5371506716. Won’t review slop.

PR #2893, opened 2026-09-14
Close HTTP response bodies on failed downloads

SamAyl:
`util.Download` returns immediately for non-200 responses before registering the deferred body close. Repeated failed plugin downloads can therefore leave response bodies and associated network resources open. Move the defer immediately after the successful HTTP request so error responses are closed too.

Adds a regression test using a tracking response body and a test transport. The test fails against the original implementation and passes with this change; all `go test ./util -count=1` tests pass. Formatted with gofmt and checked with `git diff --check`.

The commit includes the required DCO sign-off.


chadlwilson:
> The commit includes the required DCO sign-off.

Not quite: 

> Commit sha: cf1f85d, Author: SamAyl, Committer: GitHub; Expected "SamAyl sandeep.elec07@gmail.com", but got "Sandeep Ayloo sandeep.elec07@gmail.com".

chadlwilson (approved):

chadlwilson:
If you can fix the commit author and bump the version as in the bit xomment, we can merge and release this.

PR #2894, opened 2026-09-16
Release 1.6.37

areddy548:
Need new release to get the latest `adm-zip` package which got released 5 days back.

```
{
          "name": "CVE-2026-76845",
          "type": "CVE",
          "severity": "MEDIUM",
          "score": 6.5,
          "cvss3Severity": "MEDIUM",
          "cvss3Score": 6.5,
          "scoreMetadataVector": "CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:C/C:N/I:H/A:N",
          "publishDate": "Aug 24, 2026, 12:00:00 AM",
          "url": "https://www.mend.io/vulnerability-database/CVE-2026-76845",
          "description": "adm-zip 0.5.9 through 0.6.0 follows symbolic links at the extraction destination. Utils.sanitize in util/utils.js enforces containment by comparing only the string form of an archive entry name against the resolved extraction root, and Utils.writeFileTo opens the computed destination with fs.openSync(path, \"w\", 0o666), which resolves symbolic links and carries neither O_NOFOLLOW nor a pre-write fs.lstatSync check. When a path component at the destination already exists as a symbolic link pointing outside the extraction root, extractAllTo, extractAllToAsync and extractEntryTo write the entry contents through that link and then chmod its target, placing attacker-controlled content in a file outside the root without any traversal sequence appearing in the archive. Reaching the write requires overwrite to be enabled, because the preceding fs.existsSync check also resolves the link and otherwise declines. An attacker able to create a symbolic link inside a shared, reused or predictable extraction directory, such as a temporary directory or a continuous integration workspace, can overwrite any file the extracting process is permitted to write.",
          "references": []
        }


chadlwilson (approved):

chadlwilson:
Ack, thanks for this

